PR #1, opened 2015-10-08
Repository Syncing

FGDATA:
This commit brings the repositories in sync.
It brings commits from FGDATA, and FGADDon.
No changes are added to the code.

Best
@IH-COL


gilbertohasnofb:
Is this really necessary? This will bring history about things that are not in use any longer; for instance, the 3D model here was created by scratch by @HHS81 and I don't like the idea of having tons of history about the 3D model which is not in use.

gilbertohasnofb:
I don't know Israel, this seems to bring a lot of rubbish with it and I really don't care about this issue. I know you like to keep history so you may do that in your fork in FGMEMBERS, but I wouldn't like this here.

@HHS81 do you have any opinions on this? If you also don't care for these ancient commit logs then we can just close this PR.

HHS81:
I ignore Israel, so I won't merge this into. Beside the fact, that the C182S is a new model and only roughly based on Stuarts C182.

Issue #2, opened 2015-12-26
re-UV-map rudder pedals 


gilbertohasnofb:
Re-UV-map rudder pedals as they are not in the same size in the texture file

gilbertohasnofb:
We can check this out once I push the textures as it will make this much easier I suppose. Just give me a couple of days.

HHS81:
Fixed Issue #2

HHS81:
Hi,

did a mistake with the pedals uvmapping, I remapped them to indentical size, but i didn't use your texture as reference. I will fix this.
The Alt Static Air Knob is uvmapped, it just needs a red knob texture. All other things will not be influenced in any way.

gilbertohasnofb:
> I will fix this.

If you want, I can do it myself, since I will make corrections to the textures anyway. So don't bother :wink: 

> The Alt Static Air Knob is uvmapped, it just needs a red knob texture. All other things will not be influenced in any way.

:+1:

HHS81:
I fixed it now to your layout, it was easier for me than I think it will be for you. And I was responsible for the mess.... ;-)

gilbertohasnofb:
hehehe thanks then, Heiko!

PR #44, opened 2016-01-29
Electrical lightsystem

HHS81:
Using the work on the electrical and lighting system as test and practice for pull requests. (never tried before)
The whole systems are stil wip, and needs some work


HHS81:
O.k, now that's easy

onox:
> and practice for pull requests. (never tried before)

I have a couple of suggestions when using PR's on Github that you might want to use:

1. Use a short descriptive message in the summary (line 1) in your commit. The rule of thumb is to keep it less than 50 to 72 chars. Optionally you might want to add an extra multi-line message. Keep line 2 blank and add the multi-line message on line 3, 4, etc. Not all applications word-wrap automatically, so you should wrap it yourself. With `git commit -s`, git will add "Signed-off-by" on line 3 for you.

    [This](http://chris.beams.io/posts/git-commit/) is a nice blog post about how to write good git commit messages.

2. You might want to set a username. Your name is currently "unknown" instead of something like "HHS81". 

1. If the PR fixes a particular issue, then name the branch something like `bug-1`. If you then merge the branch you'll get a merge commit message like `Merge pull request #2 from HHS81/bug-1`. This way it is clear which bug got merged (1) while you can keep descriptive messages in your commit. So write something like `Add instrument X` instead of just `Fixes #1`. You can refer to the issue in the body (line 3, 4, etc.) If Github is/goes down, you can still understand what the commit does.

gilbertohasnofb:
@onox got to love that xkcd comic on the top of the page :wink:

HHS81:
@onox 
Great, many thanks, will follow it. 
:smiley:

Issue #61, opened 2016-02-01
FuelQTY, VacAmp and ManifoldGauge instruments are not correct


HHS81:
all 3 instruments are not correct drawn and modelled (position of needles etc....), which is cause for #58

HHS81:
see PR #62

gilbertohasnofb:
Closed via https://github.com/HHS81/c182/pull/62

Issue #58, opened 2016-02-01
Engine gauges texture problems


gilbertohasnofb:
There are some problems with the new engine gauges, see: 

Everything marked in yellow is either cut or too close to the borders. Everything marked in purple have being UV-mapped into two different heights which can be seen from some angles (look at the PSI text in the image for instance):

![fgfs-screen-001](https://cloud.githubusercontent.com/assets/5700795/12719594/53033f64-c8f6-11e5-9da4-4f5756b00dd3.jpg)

HHS81:
Please see also #61

gilbertohasnofb:
https://github.com/HHS81/c182/pull/60 has fixed part of the issue, so now it's missing the label FUEL in the fuel gauge (which I didn't circle above but I think it's too close to the border) and the VAC AMP gauge.

HHS81:
see PR #62

gilbertohasnofb:
Closed via https://github.com/HHS81/c182/pull/62

PR #62, opened 2016-02-01
-Fix Issue #61

HHS81:
also:
-improved tach
-improved hobbs meter

Signed-off-by: HHS81 <Heiko.H.Schulz@gmx.net>


gilbertohasnofb:
@HHS81 I will check this now. By the way, it's always good to list all issues in the PR description, such as in:

Fixes #61 and #58

This not only keeps it organized, but (normally) closes the issues automatically once the PR is merged!

gilbertohasnofb:
Also, the title "Fix Issue 61" isn't really good as it forces anyone interested in then chasing what is the issue 61 about. Something like "Fixing engine gauges" would be better. It is also good if you list everything that is improved, so in this case:

- fixing textures of FuelQTY, VacAmp and ManifoldGauge
- fixing position of the needles for FuelQTY, VacAmp and ManifoldGauge
- improvement to the tachometer and hobbs textures

Or something like that :smile:

HHS81:
O.k. thanks!

gilbertohasnofb:
No problems!

Issue #56, opened 2016-01-31
switches are not animated


HHS81:
The switches on the panel are not animated and textured.

gilbertohasnofb:
Let me know if you need textures for them.

HHS81:
I wonder where I should put them on the uvmap

gilbertohasnofb:
There is still some tiny places in the main panel texture (such as in the holes of the engine gauges). Since all switches will have the same metallic look (right?) we can simply add a single switch to one of those holes.

HHS81:
How should I uvmap them?
looking directly on it, or from side view?

gilbertohasnofb:
Probably top view (then mirror the texture on the bottom). See for instance the texture used in the c172p: 

![untitled](https://cloud.githubusercontent.com/assets/5700795/12706023/2909e4d0-c87c-11e5-83b1-0d4f21d1486e.jpg)

HHS81:
Thanks!

gilbertohasnofb:
My pleasure!

gilbertohasnofb:
Heiko, what do you think if we do things a bit differently this time: what if I just add this switch texture to the file, and then you use it as source for mapping? Would that work for you?

HHS81:
Yep, that would be perfect!

gilbertohasnofb:
Ok, it's done! As for the liveries, I will check that paintkit in the next days as soon as I have some time

HHS81:
see PR #65 for review

Issue #43, opened 2016-01-29
Electrical and lighting system


HHS81:
**Update 08/19/2016:**
Enable the domelights (high priority)

Use the electrical system from the SenecaII and adapt it according POH, add battery behavior at cold temperatures. Advantage: realistic battery behavior, lights dims when Starter is is used, and other realistic things when battery state is low (middle priority)

**Idea: 01/29/2016**
-updating the elctrical system according POH
-updating lighting system according POH
-how to render courtesy light in the lower surface of wing (already painted):
- lightmap 
  or
- emitting plane

??

HHS81:
Update: 
create one lightmap affecting panel and pedestal:
- glareshieldlight
- pedestallight
- maybe the two domelights

create lightmap for forward cabin section:
-domelight left
-domelight right 
-maybe the glareshield light (if it affects the the forward cabin)

create lightmap for the rear cabin
-domelight rear

fall back with the old fashioned material animation for non-shader users.

HHS81:
![glareshield_pedestallight](https://cloud.githubusercontent.com/assets/12649375/14804505/59e4b99c-0b64-11e6-87df-334f4e0bbf6b.jpg)
![rearcabinlight](https://cloud.githubusercontent.com/assets/12649375/14804508/61a9c0e6-0b64-11e6-811a-4ed3280d1104.jpg)

gilbertohasnofb:
These look amazing!!

As for the instrument illumination, are they supposed to be _that_ yellow in the 182? I have the feeling we have to tone down the saturation of the instrument lights quite a bit.

HHS81:
> As for the instrument illumination, are they supposed to be that yellow in the 182? 

It is difficult to match the real lighting, but yes, it might be a bit to yellow. But there is another thing, which shows the limits of lightmaps. The glareshieldlight, with this fluorescent lighting of the panel also lighten the instrument faces. So the illumination should then change more to a white. Needs maybe a lite nasal magic....

HHS81:
Further work:
![fgfs-screen-014](https://cloud.githubusercontent.com/assets/12649375/14927897/f9249ba6-0e54-11e6-8b6c-91d29e813734.jpg)
![fgfs-screen-015](https://cloud.githubusercontent.com/assets/12649375/14927898/f9288a2c-0e54-11e6-9f15-d3977771e3a8.jpg)
[...]

PR #65, opened 2016-02-02
Issue #56: animation and texturing switches

HHS81:
textured switches
animated switches
corrected Master switches as they had been assigned to wrong functions
animated domelight buttons

Signed-off-by: HHS81 <Heiko.H.Schulz@gmx.net>


gilbertohasnofb:
Beautiful work, thanks a lot!

gilbertohasnofb:
Hmm, there is a conflict with this branch. It seems that `dev/C182S.blend` has been modified on this branch, but it also has been modified on the previous PR about the flaps :cry: I am not very familiar on how to solve this type of conflicts, do you mind if I ask onox for some help?

gilbertohasnofb:
And by the way, now that the radio and instrument illumination is working, how will we handle the panel illumination? Are we going to use a light map?

HHS81:
> Hmm, there is a conflict with this branch. It seems that dev/C182S.blend has been modified on this branch, but it also has been modified on the previous PR about the flaps :cry: I am not very familiar on how to solve this type of conflicts, do you mind if I ask onox for some help?

I will look into, the `dev\c182s.blend` here in the branch should be the newer one. No idea why there is a conflict.

> And by the way, now that the radio and instrument illumination is working, how will we handle the panel illumination? Are we going to use a light map?

Yep, but before I want to alter the avionic illumination. They all behave completely different now, and all aircraft in fgfs and in X-Plane even handle it all in a different way. Some illuminate the face, some just the LED-numbers, some only the buttons....

After reading the POH, it seems to me the radio-light-button dims the LED-numbers and the button. 
The face are iluminated by the glareshield-light. 

I have written up what I plane in Issue #43

gilbertohasnofb:
> I will look into, the dev\c182s.blend here in the branch should be the newer one. No idea why there is a conflict.

The conflict is because the last PR also had a modification to that blend file. So basically we had the blend file in the master at a stage A and two PR's with the blend file at stages B and C respectively. When the first PR was merged, the master's blend file went from A to B as instructed. But the second PR wants to change that blend file from A to C, and not from B to C. This is what is causing the conflict, it simply doesn't know what to do. Have you intended to modify the blend file in that flaps PR? If not, then probably you created the new branch while you already had modifications to the master, which may be the cause of the conflict. Makes sense?

And thanks for the info about the illumination!

HHS81:
make sense, yes. Now, how to solve it?

gilbertohasnofb:
The lazy way is to cancel this PR and delete this branch. Then create a new branch using the current master, and then reapply the modifications (basically recopy the files). Alternatively, there is some way of solving git conflicts via the terminal by choosing which file is to stay, but I am not familiar with that at all. If you want to do it the proper way, I'd have to ask help to onox or someone else from the c172p project.

HHS81:
I use the lazy way, it is already very late here. Solving via terminal don't work, I tried.

gilbertohasnofb:
Okay. So absolutely always before creating a new branch you should do a `git status` to confirm two things: 1) that you are in the master branch and 2) that it is clean. Also, always do a `git pull` before as well, to make sure that the master is the most current.

HHS81:
next try - no error messages yet. Wait for your review

gilbertohasnofb:
:+1: I will do it right now

Issue #68, opened 2016-02-03
Make use of procedural lights


HHS81:
Make use of procedural lights for:

- Navlights red, green, tail
- Strobe

- Beacon (if offset can be defined)

not for landing or taxilight - I like those textures :heart_eyes:

Test available at: [https://github.com/HHS81/c182/tree/newlights](url)

gilbertohasnofb:
Great, I will test them soon :smile: (and yes, your landing and taxi lights are amazing!)

gilbertohasnofb:
Looks good, but I found some problems with the lights in general:

- maybe the tail light has to be moved slightly backwards as there is a black dot in its centre (is that the fuselage colliding with it?)
![fg-0](https://cloud.githubusercontent.com/assets/5700795/12790709/e51c5dfa-caa3-11e5-8081-ede5ce78c3d5.jpg)

- same is true for the strobes
![fg-3](https://cloud.githubusercontent.com/assets/5700795/12790711/e52517ec-caa3-11e5-9c27-760601aa8541.jpg)

- strange triangle inside the beacon
![fg-1](https://cloud.githubusercontent.com/assets/5700795/12790710/e51dd248-caa3-11e5-917b-193adc814114.jpg)

- strange red light above the antennas, what is that?
![fg-2](https://cloud.githubusercontent.com/assets/5700795/12790712/e524cc92-caa3-11e5-984a-8e8c94ae40cf.jpg)

- IMO, the nav lights should be stronger, I can barely see them now during the day.

HHS81:
Lol, the strange red light is the new procedural light testing the angles.
All other lights are the one we were used to. No change on that yet.
Those artifacts you mentioned I've never seen. Especially not on the beacon.

gilbertohasnofb:
Lol :stuck_out_tongue_winking_eye: :dizzy_face: facepalm...

Anyway, I have not seen those artifacts before either! I have no idea what is causing them, but if we plan to move on to the procedural lights it won't matter anyway. But it's really strange how the nav lights look so dim now, but maybe I got used to how the look on the c172p after so much testing there.

But about the procedural light: the effect is really cool, but what I deeply dislike on it is that the centre is not completely white like our png lights, but rather a shade of red. Thorsten actually made this modification to all lights, and I think they look so much worse then before. But I remember some people complaining they couldn't make the colour of the PAPI lights out, so I can't completely blame him. But it would be great if we can tweak these, as the png lights are still looking better IMO.

[...]

PR #70, opened 2016-02-08
Issue #68

HHS81:
Issue #68 Make use of procedural lights

-  added procedural lights
-  use of correct directional angle
-  made procedural lights selectable dependant on renderer (ALS) and user choice via Aircraft-Menu
-  changed electrical.nas to make the new lights working


gilbertohasnofb:
Okay, so I tested this and I have just one comment: I know you don't like the star effect, but to leave the nav lights and the beacon without the desaturated centre (which causes the star effect) makes them look so unrealistic in my opinion, they look just like a plain blob of colour, not like a light source to me. The strobes are looking really nice and the tail light as well.

I also found a little bug unrelated to the procedural lights but you might want to take a look at it in this same PR. Basically the landing and taxi lights look buggy from certain angles:

fgfs-screen-001

But if I turn the camera a tiny bit they look normal again:

fgfs-screen-002

wlbragg:
>but to leave the nav lights and the beacon without the desaturated centre (which causes the star effect)

How much desaturation is needed before the star effect kicks in? Is there a middle ground that will stop the star effect but still allow enough desaturation to look good?

gilbertohasnofb:
You know, in the end we are just trying to hack Thorsten's effect. Probably the best approach would be to ask him to reconsider it if we all feel strong about it...

HHS81:
> but to leave the nav lights and the beacon without the desaturated centre (which causes the star effect) makes them look so unrealistic in my opinion,...

making the centre white makes the color disappear and you have a white star...., or a grey with a bit color an the outer part ....
But the color must be clear visible, otherwise no one recognize the flight direction of the aircraft.

Sorry but the lights simply not behaves like the Gimp SuperNovas. That's the best thing I can do - I don't like it:
fgfs-screen-001

fgfs-screen-002

I prefer to recognize the colors,

gilbertohasnofb:
> But the color must be clear visible, otherwise no one recognize the flight direction of the aircraft.

Absolutely!

> Sorry but the lights simply not behaves like the Gimp SuperNovas. That's the best thing I can do - I don't like it

Look, I don't think the result is bad, I just think those textures were looking much better. But that said, there are other good things from a procedural approach.

Maybe someone wants to show those old lights we had to Thorsten and see if he would be up to try to "recreate" them procedurally.

HHS81:
Just give him that: 
_Light effect decreases according to 1/r where r is the distance from star center._
_Radius
This is radius of the SuperNova center (1-100). When you increase the value, you increase the number of central white pixels according to r*r (1, 4, 9...)._
_Spikes
This is number of rays (1-1024). Each pixel in the nova center emit one pixel wide rays._

from: https://docs.gimp.org/en/plug-in-nova.html

HHS81:
And ask him, which effects he does  likes on this pircture....:
fgfs-screen-007

HHS81:
New try - please review

HHS81:
@gilbertohasnofb  Ready for review!

gilbertohasnofb:
Really nice work, Heiko! Looks much better IMO. At night-time, it looks fantastic!

HHS81:
Thanks!

Issue #55, opened 2016-01-31
Idea/ Concept: place chocks/ safetycones/ fuel-trucks etc. via model-manager


HHS81:
Place the chocks/ safetycones via model-manager like the walker

Advantage: 
- the model acts like a scenery model, will be recocgnized as collision object - no need for a workaround with the parking brake

- safety cones and other objetcs stays where they are placed (no flying trucks....)

Challenge: nasal script and making the objects remove from scenery  when not needed

Basic Idea: Walker

HHS81:
Update:
took the code from the walker - and it works!! :smile:
The wooden-chocks.ac holds the aircraft at its place - even with full thrust!
The wooden-chocks keeps their position- even if the aircraft is moving. No flying trucks.... :yum: 

![nobrakes](https://cloud.githubusercontent.com/assets/12649375/13024766/9c0a3ed0-d1fa-11e5-9118-bc147c88289d.jpg)

Push soon

HHS81:
Known Issue: 
the point of the ground reaction is exactly at the middle bottom of the wheel - that's explain the offset to the chocks in the picture. 

- FDM-wheels needs probably one more contact point?
- or a invisible collision object with an offset?

Guess the later makes more sense, the first one is more realistic but might be problematic on the normal ground....

HHS81:
Known current Issue as well:
- possible interaction with a later added walker - have to test it

gilbertohasnofb:
I'd say the yellow chocks would fit better, the idea of the wooden one is to look "improvised", just like one can expect to find in a bush plane in the middle of Alaska for instance :smile: 

As for the fuel-trucks, I don't think it's common to refuel these single engine planes with trucks, but rather using a gas pump in the airport, though I may be mistaken here.

HHS81:
Hi,

yep, the yellow chocks are indeed better. 
The truck was just mentioned as an example. The whole thing is easy to implement, there is no bad interaction with the walker and no fps-impact. 
Maybe something for the c172p...

I found that today: http://www.aircraftspruce.eu/images/products/Airplane_Ready_to_Refuel_Small.jpg
Maybe?

The Cessna 182 allows external power supply, so at least a small ground power unit will be implemented, like I did for the EC 135 P2.

gilbertohasnofb:
Sounds all good, but about the small refuel unit, how much fuel does that carry? Can you really fill two tanks with it?!

HHS81:
[...]

PR #77, opened 2016-02-14
Issue #55: Static elements

HHS81:
Fix Issue #55 and Issue #82
updated 04/24/2016:

- use of geo.put_model for dynamic adding objects into scenery
-  each wheel chock clickable (with clickspots)
-  each clickable safety cones
- selectable wheel fairings
- changed color of tie-downs to yellow 
-  add walker
-  ladder (right or left), compatible with walker
-  external power unit
-  portable trailer fuel tank
-  implementation of aircraft tie-down
- gui update which reflects changes above
- changes to fdm with correct aileron deflections; forward slip and with that much more realistic crosswind behavior
- ASI reacts correct on failure/ pitot cover
- external power use of electrical system
- automatic disconnect when aircraft is moving

and maybe some minor things.....


HHS81:
P.S.: changed key-binding for engine starter (now shift-s) due to walker.

gilbertohasnofb:
I will review this now, but I strongly advise not to change the binding of the s key for the engine starter, that's the default for all aircraft and there will be tons of complaints like "why engine no work FG sux!" in the forum. Please reconsider it.

HHS81:
> I will review this now, but I strongly advise not to change the binding of the s key for the engine starter, that's the default for all aircraft and there will be tons of complaints like "why engine no work FG sux!" in the forum. Please reconsider it.

Wait a moment pelase,....I think I know what happened.....

gilbertohasnofb:
ok

HHS81:
No, sorry.
There is strange bug with the walker. The s-key is only working in walker-views. It should be the way around.
I don't understand why. The author of the walker is already informed, I wait for answer. 
So the key-change is only temporarily, and mentioned in the help-file.

HHS81:
We can change the key later, so if you want merge it though, I would be happy.

gilbertohasnofb:
Okay, some comments:

- I still think the shift+S change is bad... we are kind of "breaking" something to revert it later
- `take ladder` toggle: is it really necessary? Why not only `place ladder on the <x> wing`?
- cones are not "clickable"
- pitot tube cover doesn't kill AIS
- tail tie-down: I'd rather have an shallower angle such as this:
fgfs-screen-001

What do you think?

HHS81:
O.k., the electrical.nas is the cause for the issue. Why ever.... needs some time to solve.
ladder: I think about how to get this like proposed
cones: should they be? They are not part of the checklists.
pitot cover killing IAS: later together with pitot-icing
tail-tie-down: o.k.

So wait with the PR.

gilbertohasnofb:
Ok Heiko, it's quite late here so I am going to sleep :zzz: I can merge this tomorrow when we are ready, ok?

> cones: should they be? They are not part of the checklists.

Yeah, I see your point, but I just don't see why not. Personally, I only use dialogs when I am forced to, I think they really break the immersion. Anyway, this is a minor detail.

Great work today, by the way. Cheers!

HHS81:
s-key problem solved.

HHS81:
ladder dialog and tiedown angles changed.
Everything else this or tomorrow evening

gilbertohasnofb:
> s-key problem solved. 
> [...]
> ladder dialog and tiedown angles changed.
> Everything else this or tomorrow evening

Great, Heiko! Really appreciate the effort :+1: So when you are ready with the corrections, let me know so I can take another look at it and then merge it!

HHS81:
@ onox: I guess it is an extended version of the nasal/math.nas in FGData?
Wouldn't be it better to get math-ext.nas into FGdata/nasal?

The tiedown is a great feature, but honestly said a bit pain to implement. Would be great if there would be a more simple method to include this (similar like the pushback truck) :smiley:

HHS81:
> ladder dialog ....
FAILED!

the solution I choosed adds with every click on right or left ladder another instance model[0]; [1], [2];...
Without adding another 80lines of nasal-script I can't get it work.

The same would apply to the safety-cones and the chocks. 
Sorry, no idea how to do it better.

gilbertohasnofb:
No problems then, Heiko. I think that clean GUI are very important, but if there is a practical reason or a limitation behind this decision, then we can live with it. Maybe one day you will think of an way of solving it :bulb:

HHS81:
Maybe. maybe using geo.put-model() is indeed the better solution.....

HHS81:
This is all I can think:
Enabling makes the ladders selectable
Unfortunately I can't make them seperatly working like that as clickable. 
That applies to chocks and pylons as well

So no real solution, beside geo.put-model() just for keeping the code small
fgfs-screen-006

gilbertohasnofb:
Ok Heiko, I am not familiar at all with the code behind these. But I'd gladly test and merge this once you are happy with this PR, even if the solutions aren't perfect yet :smile:

HHS81:
@gilbertohasnofb reade for review, please see entry of this PR for a complete list of changes

gilbertohasnofb:
@HHS81 Ok Heiko! So I had a look at it, here are some thoughts:

- left and right tie-down can be removed by clicking on them, but cannot be added by clicking on the hotspot
- tail tie-down can't be added nor removed by click, only via menu
- chocks were a bit buggy sometimes. Adding them make the aircraft jump a bit in the air. Other times the wheels were completely not touching them (though this seems to be a bit more rare). Overall they work, but there is a lot of potential for bugs IMO.
- is it possible to disable wheel chocks and cones hotspots while on the air? It looks kind of ugly when pressing Ctrl+C now
- clicking on cone hotspots in the air toggle them between ON and OFF, which is strange
- automatically disconnect power unit if groundspeed > 0.0? Right now it looks funny if you forget the cable connected then start the plane

What do you think?

HHS81:
@gilbertohasnofb 

> left and right tie-down can be removed by clicking on them, but cannot be added by clicking on the hotspot
tail tie-down can't be added nor removed by click, only via menu 

I improved the clickspots by simplifying it, so you should be able to add and remove them by clicking on it, unless the aircraft is moving

> chocks were a bit buggy sometimes. Adding them make the aircraft jump a bit in the air. Other times the wheels were completely not touching them (though this seems to be a bit more rare). Overall they work, but there is a lot of potential for bugs IMO.

I tried to improve it. Nethertheless it will only work o.k., when parking brake is set. That's how it will work in real life as well - the chocks are [only a safety addition to the parking brake](https://en.wikipedia.org/wiki/Wheel_chock). 
Also the jumping depends on the aircraft wheight. Less or even no jumping when aircraft is empty. JSBSim weirdness....

> is it possible to disable wheel chocks and cones hotspots while on the air? It looks kind of ugly when pressing Ctrl+C now
clicking on cone hotspots in the air toggle them between ON and OFF, which is strange

Done by disabling the clickspots when aircraft is moving

> automatically disconnect power unit if groundspeed > 0.0? Right now it looks funny if you forget the cable connected then start the plane

Good idea, done that way. 
But more realistic would be if distance >0.0. Will be a task for Milestone 2: I have the code somewhere for it since I used something similar for moving helipads, but can't find it here on my laptop. Not a showbreaker I think.

gilbertohasnofb:
I will check this now, thanks for the quick fixes!

gilbertohasnofb:
Heiko, everything is much better!:

- hot spots are all good now, but I think you had forgotten to add the speed condition to TiedownHotSpotLeft and TiedownHotSpotRight, and so it was possible to activate the tie-downs even if the aircraft was at a huge speed. I added these myself.
- The Fuel Tank Trailer label was spelt altogether in the menu (i.e. "Fueltanktrailer"), so I fixed that as well
- chocks seem much better now! I know that you are supposed to use them with parking brakes and engine off only, but I always try to "break" the plane while testing to find potential bugs. 

> But more realistic would be if distance >0.0. Will be a task for Milestone 2 [...] Not a showbreaker I think.

Sounds good, and definitely not a show breaker! :smile: I think we just need to solve the issues with the engine startup status and then we are ready for FGADdon!

gilbertohasnofb:
So, just confirm if you are okay with my modifications in this branch, and we can merge this PR!

HHS81:
> hot spots are all good now, but I think you had forgotten to add the speed condition to TiedownHotSpotLeft and TiedownHotSpotRight, and so it was possible to activate the tie-downs even if the aircraft was at a huge speed. I added these myself.

Oh, indeed. Thanks.

> The Fuel Tank Trailer label was spelt altogether in the menu (i.e. "Fueltanktrailer"), so I fixed that as well

yes, thanks. I accidently "germanized" it.

> chocks seem much better now! I know that you are supposed to use them with parking brakes and engine off only, but I always try to "break" the plane while testing to find potential bugs. 

Yes, I like this method to test. Meanwhile I had an idea, why it sometimes jump. It seems that the contact points are not matching correctly, due the gear compression in jsbsim. I have to think about a method how to let them match perfect. But this maybe for a later milestone, nowI think it is good enough. 

So I conform I'm o.k. with your modifications! :smile:

gilbertohasnofb:
>  It seems that the contact points are not matching correctly, due the gear compression in jsbsim. I have to think about a method how to let them match perfect. But this maybe for a later milestone, nowI think it is good enough. 

Indeed when the bugs would appear the gear would change compression quite a lot up and down, so maybe we can have a look at it later. But let's merge this now! :smile:

gilbertohasnofb:
:clap: :clap: :clap: very nice work, Heiko! I particularly love how the walker can climb the ladder, it's really amazing :smile:

HHS81:
Yep, that was the reason why I wanted to use this method :smile: 
Thanks for merging, this was a long work, but worth it!

gilbertohasnofb:
The result is really amazing, I am very happy with it!

PR #90, opened 2016-02-16
implementing random splash screens + new splash and thumbnail

gilbertohasnofb:
Implementing random splash screens + new splash and thumbnail

Closes https://github.com/HHS81/c182/issues/49


gilbertohasnofb:
:warning: :warning: :warning: do not merge this yet! I noticed that splash4 uses a cockpit image that you posted with a nice AOMap, but it's missing the engine instruments. Would you like to take a new shot of the cockpit so I can create the splash screen and we can merge this, Heiko?

gilbertohasnofb:
Heiko, no pressure but do you have some opinion on this? If you don't feel like messing around with these I can try to take a nice screenshot of the panel to replace `splash4`, just let me know.

HHS81:
do we need a splash screen of the cockpit? I would like to create a new AOmap and then add this, but it needs a while.

gilbertohasnofb:
Personally, I think it's good to have a shot from the cockpit, particularly since it looks really good IMO :smile: But in that case, I think I will simply take a new screenshot for splash4 so we can merge this, and then once the AOmap is done we can replace it.

HHS81:
> Personally, I think it's good to have a shot from the cockpit, particularly since it looks really good IMO :smile:

Yes, that's true. But I do think, that people will be confused seeing a splash screen showing something, which looks different then. 
I'm currently busy with real life as I have a job as a research assistant at an university hospital, so in the moment things slowed down a bit.  

So if you want to take a new splash screen, that would be o.k. and I would merge the whole thing. 
Until now I wasn't aware of this feature of random splash screens, and I think that it is very cool! So we still can add other and more splash screens later!

gilbertohasnofb:
> Yes, that's true. But I do think, that people will be confused seeing a splash screen showing something, which looks different in the sim then. 

Yes, you've got a point here. So I replaced splash4 with an external screenshot, and so this is ready to merge.

> Until now I wasn't aware of this feature of random splash screens, and I think that it is very cool! So we still can add other and more splash screens later!

It's a very cool feature indeed, I first came across it with the c172p. So we can add more later for sure!

HHS81:
Done!

gilbertohasnofb:
Thanks! :+1:

Issue #71, opened 2016-02-08
Wrong Power and FuelFlow settings 


HHS81:
Currently power and FuelFlow were set to match the values in POH. But the power values are not matching with the POH.

Problem:
Accidently I used the wrong mixture setting.
Cessna recommends Peak EGT, I used to trim the values with Best power-mixture (125°F Rich).

Solution:
retune the values for cruise perfomance with the right mixture setting PEAK EGT!

HHS81:
Mainly fixed, but wrong horsepower at 4000ft @ 24 MAP (but correct at 23 MAP) so keeping this open until I fixed it.

PR #91, opened 2016-02-28
Fdm #85 and #71

### LINKED ISSUE #85: strong pull to the right during take off roll
I always feel a strong pull to the right during take off roll and often I am applying near to full rudder to keep the aircraft in the runway. Once the aircraft leaves the ground, it immediately tries to go to the left (given the near full rudder). Is this behaviour normal/intended?

HHS81:
fix #85 and #71 
- using all real values by David Megginson
- allowing forward-slip and with that crosswind landings at higher wind speeds than 15ktn
- stable take-off behavior with less pronounced roll - but probably still the [same issue like c172p](https://github.com/Juanvvc/c172p-detailed/issues/721)
- fix of inverted animation of rudder trim in cockpit
- small changes on stallhorn.xml


gilbertohasnofb:
:+1:

PR #95, opened 2016-03-27
ALS landing and taxi lights

gilbertohasnofb:
ALS landing and taxi lights

Closes https://github.com/HHS81/c182/issues/94

### DIFF
--- a/Systems/als-lights.xml
+++ b/Systems/als-lights.xml
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+    Copyright (c) 2015 onox
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


gilbertohasnofb:
@wlbragg do you know if the ALS landing lights are supposed to illuminate the propeller when the engine is off? Is there a way of making it illuminate only the ground?

wlbragg:
Wow, I never noticed that one. I'll have to take a look at it and see what I think.

gilbertohasnofb:
Yes, I tested it in the c172p and it also has some illumination in the propeller as well, but it's somehow less than the c182 (probably due to the viewing position and model size). If you think of something that could work, let me know.

wlbragg:
I was able to get it off the prop using landing-light-offset3-deg set to 10.  landing-light-offset2-deg had an effect also. I don't recall of the top of my head which does what.

gilbertohasnofb:
offset3 is the vertical offset, offset1 is the lateral offset. I found out that moving the lateral offset to -3 does the trick for the c182, so thanks for pointing it out!

gilbertohasnofb:
@wlbragg in case you are wondering, this is what those properties do:

- `landing-light1-offset-deg` controls the landing light lateral offset
- `landing-light2-offset-deg` controls the taxi light lateral offset
- `landing-light3-offset-deg` controls both landing and taxi lights vertical offset

gilbertohasnofb:
@HHS81 this is ready to be reviewed and possibly merged

PR #98, opened 2016-04-09
better top panel texture

gilbertohasnofb:
Closes https://github.com/HHS81/c182/issues/88

Since we don't have enough resolution to get a convincing leather pattern, I tried at least to get some nice tone to fit with the rest of the cockpit (not too bright but not too dark). See:

fg

IMO it's better than what we currently have.


HHS81:
Thanks, I will take a look in the enxt days, as I'm currently quite busy. 
Sorry again for the bad uv-mapping of the panel. Will keep that in mind for future projects.

gilbertohasnofb:
No rush Heiko, whenever you have time. And no worries about the panel, your uv-mapping has been excellent in this project and the top panel isn't bad at all, it's more a matter of resolution rather than mapping itself. It's a common problem though, since very few people foresee that due to the size of this top panel one would need a resolution as large as the panel itself to get the same amount of detail. the c172p suffers from the same problem, for instance.

gilbertohasnofb:
And by the way, we could really think about asking for this plane to be added to FGAddon soon. I checked all issues marked with the milestone "FGAddon Release 1" and it seems to me that we could easily push them to "FGAddon Release 2" and work on them in the future. What do you think?

HHS81:
@gilbertohasnofb  I would like to have the StaticElements-branch complete at least and then pushed, as there are some important things inside. All other things could be in FGAddon Release 2.

HHS81:
@gilbertohasnofb  yeah, you are right with the top panel/ glareshield. Something I have to remember with future projects.

gilbertohasnofb:
First, thanks for reviewing this! :+1: 

> I would like to have the StaticElements-branch complete at least and then pushed as there are some important things inside. All other things could be in FGAddon Release 2.

Fine with me. So once we have that branch merged we can contact the devs for a FGAddon merge. To whom shall we speak to exactly?

HHS81:
Stuart Buchanan is currently the maintainer of the C182 in FGAddon, as much as I know. We probably need to make sure the engine is running when FGFS is launched, as a current discussion on the list is pointing out. Btw,  started by David Megginson - the one who started the c172 and c182 a long time ago, and provided us with nice fdm-datas. :smiley:

HHS81:
Btw.: I have commit rights for FGAddon. :smile:

gilbertohasnofb:
> We probably need to make sure the engine is running when FGFS is launched, as a current discussion on the list is pointing out

Yeah, I have seen that but honestly I am not convinced by that, though I don't have a strong opinion on it either. The only thing is if we will add this "feature" we must be create an option in the menu to deactivate it, something like `[ ] start with engines off` or anything like that, and make this property persistent. 

> Btw.: I have commit rights for FGAddon. :smile:

Great! So we can speak to Stuart about it (since he has been maintaining the c182) and then merge it. By the way, will he want to keep his c182 or will this project here become a replacement for it?

HHS81:
The only reason behind the whole discussion is, that parking places and short-holds doesn't work in FlightGear. I admit that David Megginson has a point when he says that it is not realistic to launch an aircraft without running engines on the runway. I think of multiplayer as well. But we wouldn't have the problem if parking-places and short-holds would work for everyone.

I couldn't talk with Stuart about the c182 yet, so I have no idea. But I would like to see a complete replacement.

gilbertohasnofb:
I agree with you and with David, the argument of engine off at a runway is very valid, so indeed we can do as you say. But I really think that we need to add an option (which is then saved for the next session(s)) in which the user can decide to start with engine off. Even if parking positions aren't working right now, you can use coordinates to start at a specific place.

> I couldn't talk with Stuart about the c182 yet, so I have no idea. But I would like to see a complete replacement.

Yes, I also think a replacement would be best, the other c182 doesn't look good at all.

HHS81:
@gilbertohasnofb  I'm in favor of an option for cold start as well, even a simple singe engine aircraft has its own procedure, and this should be covered realistic (O.k., needs fuel pump simulation though...).

gilbertohasnofb:
So how should we implement this, Heiko? The c172p has a property `/fdm/jsbsim/running` to enable/disable the option of starting the sim with engine running, as well as another one `/fdm/jsbsim/complex` which enables/disables the complex startup procedures (requiring priming, specific throttle settings, etc.).

HHS81:
I thought a bit about that. 

Imagine an new user launches the aircraft on the rwy: engine is running and that's o.k.
Now imagine another user launches the aircraft on aprking post: engine is running and that's probably not what he wishes for....

I'm currently not convinced that one or any other methos will satisfy everyone. I think in FlightGear's launcher you should be able to set the aircraft status "`running`" or "`cold start`". 
Not sure how it can be done for all launchers we have.....

gilbertohasnofb:
> Now imagine another user launches the aircraft on parking position: engine is running and that's probably not what he wishes for....

This has been my exact thoughts. Even if parking positions aren't available right now, people use coordinates to put the aircraft away from the runway for serious flights.

But in my view this is not so much of a problem if we make things persistent: if you can select which option you want in the aircraft menu, it means you will potentially have to load the sim just once before setting it _forever_. This is the behaviour of the c172p, and after I activated the complex procedures and oil management, and set the engine to NOT be running on startup, it has been perfect for me.

It of course would be cool to have a launcher option, but that's more complicated as we have tons of launchers and tons of aircrafts, and it's all very decentralized... IMO it's way easier to solve this from the aircraft development point of view than to try to add something to let's say the Qt5 launcher, just to have FGRun users complaining that the engine startup option isn't what they want.

Issue #46, opened 2016-01-31
Adding AOMaps on interior textures?


HHS81:
Should we add AOMaps on the interior-textures?
The mainpanel as an example get then some some touch like Carenado Aircraft:
![image](https://cloud.githubusercontent.com/assets/12649375/12699346/451d42ee-c7b8-11e5-95c8-fb9e141ae479.png)

Ours with 80% Opacity of AOMap (can be increased, and the whole AOMap improved by changing settings wihile generating the AOMap in Blender)
![fgfs-screen-005](https://cloud.githubusercontent.com/assets/12649375/12699423/b039a070-c7ba-11e5-8a22-0a50399bb08f.jpg)

See  AOMaps-branch

gilbertohasnofb:
Sorry Heiko but what is an AOMap and what does it do? Nevertheless, that shot of yours look stunning!

HHS81:
AO stands for Ambient occlusion: https://en.wikipedia.org/wiki/Ambient_occlusion

Though this is not quite right, then what I did and plan is more or less ambient shadows generated by Global Illumination (which actually will not conflict with Rembrandt shadows or the ALS shadows). 

I will see if I can enhance this more, so we can close to those commercial products (As much as I know some Carenado-devs uses Blender for their work....)

gilbertohasnofb:
I see, that looks really interesting!

gilbertohasnofb:
It seems that the files `dev/AOPanel.png` and `dev/AOMap.png` have wrong proportions (both are 2048x2048 while the panel and livery files aren't). I guess this won't affect anything, but for instance the AOPanel is twice the size it should be.

HHS81:
I know, it happens when I create them in Blender. Those files went accidently in the wrong branch and aren't to be considered ready to use.

gilbertohasnofb:
@HHS81 in the c172p, @wlbragg has implemented the irradiance effect which looks really good and may be related to this, since it drastically reduces the brightness of the cockpit in general when the sun isn't directly shinning upon a surface.

Also, the same effect can be used to generate some irradiance from the instrument lightning and radio digits, see: https://www.youtube.com/watch?v=o4zyCCea7Nw

And these are the PR where the effect has been implemented:
https://github.com/Juanvvc/c172p-detailed/pull/730
https://github.com/Juanvvc/c172p-detailed/pull/731

wlbragg:
[...]

PR #100, opened 2016-06-14
Issue #46: AOMap

HHS81:
adding a gimp-post-processed shadowing layer of 70% Globa Illumination and 30% Ambient Occlusion made with Blender Cycles


HHS81:
@gilbertohasnofb: 
Please review and merge.

Erik Hofman suggested in a private mail that we get the aircraft now into FGaddon, and let us fix the other issues later.
I agree, as the lightmaps needs some investigation from side to get the noise removed. This will need some time to figure out.

Do you agree?

gilbertohasnofb:
It looks absolutely incredible! :astonished:

gilbertohasnofb:
> Erik Hofman suggested in a private mail that we get the aircraft now into FGaddon, and let us fix the other issues later.

:+1: Absolutely! Let's make a release here in our repo, and then ask him to add it to FGAddon, what do you think?

gilbertohasnofb:
@HHS81 hmm, I just spotted some small problems in the AOMap for the cockpit, see how these circuit breakers and switches are super dark while others aren't:

aomap

HHS81:
Yeah, that was the reason I didn't made a pull request before. Forgot about that. Easy to fix, since I need only to exclude the switches, circuit breakers and other things from the layer in Gimp.

gilbertohasnofb:
Ok! Once you create the new PR with the fix I will test it and merge it! Let me know if I can be of any help.

Issue #92, opened 2016-03-01
create ALS exterior shadow


gilbertohasnofb:
create ALS exterior shadow

HHS81:
@gilbertohasnofb 
![fgfs-screen-005](https://cloud.githubusercontent.com/assets/12649375/16710392/82db3576-462c-11e6-9467-ca52bad1fe26.png)

gilbertohasnofb:
@HHS81 :clap: :clap: :clap:

PR #104, opened 2016-07-09
Issue92: adding exterior shadows

HHS81:
adding volume-shadows and glass-rain effect.

Waiting to be merged


gilbertohasnofb:
Very nice. The shadows are perfect. As for the rain, it's also good, but I think we can make some adjustments. For instance, the rain in the window was completely static when the airplane was not moving _but_ under heavy rain (so in RL the drops would be moving, I suppose). wlbragg has done a fantastic job with the windshield in the c172p, maybe he would like to give us a hand here?

gilbertohasnofb:
I will point this branch to him. If he doesn't have the time, then we can just merge this now and rework the effect later.

wlbragg:
I just now saw this and pulled your branch. I just fired it up to see the effect and what you have visually. Now I'll look at the code and see what you did.
In the c172p we used prop rpm and airspeed I think.
The rain should show the drops hitting the screen when the aircraft is static, I'm not sure why it's not (haven't looked at the code yet). But to try to get the raindrops to roll down like they would from just a heavy rain, that may not be possible because the shader code doesn't have that effect and to get that type of movement you would have to do it in xml or nasal like we do for the airspeed and prop rpm.
It's been awhile since I experimented with the capabilities but it is limited.

wlbragg:
Looks like you have the same code we uses in c172p, right? Then there must be a glitch somewhere, still looking.

wlbragg:
Problem is our code uses engines/active-engine/rpm you need engines/engine/rpm. That is one issue.

wlbragg:
I think I found the other problem. In c172p we check (using nasal) if we are on water to also use the splash/rain effect. We use the following code,
~~~xml
# Check if on water
var poll_surface = func
{
    var engine_rpm = getprop("fdm/jsbsim/propulsion/engine/engine-rpm");
    var hydro_active_norm = getprop("/fdm/jsbsim/hydro/active-norm");
    var ground_splash_norm = getprop("/environment/aircraft-effects/ground-splash-norm");

    # Use engine RPM and speed to control ground splash if on the water
    # and below 2 meter AGL
    if (getprop("position/altitude-agl-m") < 2 and hydro_active_norm > 0) {
        var groundspeed_half_pc = 0.005 * getprop("velocities/groundspeed-kt");
        var engine_rpm_almost_nothing = 0.005 * 0.065 * engine_rpm;

        var splash_norm = std.max(engine_rpm_almost_nothing, groundspeed_half_pc);
        setprop("/environment/aircraft-effects/ground-splash-norm", splash_norm);
    }
    elsif (ground_splash_norm > 0)
        setprop("/environment/aircraft-effects/ground-splash-norm", ground_splash_norm - 0.005);
}
~~~

That last elsif assigns -0.005 to ground_splash_norm if not on water. I think that is the number needed to get the sprinkles on the windshield.

I'll go ahead and tweak the necessary code to get this fixed and push to issue #92.

wlbragg:
Hum, I think I have to take that back, looks like ground_splash_norn may be a dead property no longer being used. Still looking.

wlbragg:
Hum, I can't track down why it's static when sitting there. All the numbers look correct. Still looking.

Are either of you getting this error/warning?
~~~xml
In file F:/FlightGear/Custom Aircrafts/c182/Systems/indicated-airspeed.xml: line 29
      Property /instrumentation/airspeed-indicator/indicated-speed-kt is already defined.

In file F:/FlightGear/Custom Aircrafts/c182/Systems/indicated-airspeed.xml: line 29
      Property /instrumentation/airspeed-indicator/indicated-speed-kt is already defined.

In file F:/FlightGear/Custom Aircrafts/c182/Systems/indicated-airspeed.xml: line 30
      Property /sim/model/c182/securing/pitot-cover-visible is already defined.
Failed to tie property /systems/asi/indicated-airspeed to object methods
Failed to tie property /systems/asi/pitot-serviceable to object methods
~~~

wlbragg:
I'm stumped as to why the rain is static when sitting without the engine running. It should be working. I'll come back to it later and try again. For now I'm pushing the change engines/engine/rpm to glass-rain.xml

wlbragg:
oops, guess I don't have permission to push. Anyway, @gilbertohasnofb I think you know what to change.

HHS81:
@wlbragg 
Thanks for your hard work. Really appreciated!  I have give you push permission, so you can push it now. 
And yes, I get this warnings. I knew how to fix it, but somehow I lost the solution for. Will look for it at the weekend. I'm quite busy under the week now, find only some time at the weekend.

wlbragg:
Thank.

>I knew how to fix it, but somehow I lost the solution for.

All you have to do is initialize the parameter indicated-speed-kt in the fdm section of the -set.xml file instead of adding the new property in indicated-airspeed.xml

Some would rather put up with the warning and leave the property defined in the ancillary file rather than move it to the set file. I'm the opposite, we define almost everything in the set file and use it later in other ancillary files, I think it should be defined in the set file along with everything else.

wlbragg:
It still says I don't have permission.

gilbertohasnofb:
@wlbragg thanks for all this work!!

HHS81:
@wlbragg Thanks, that it is much bettter! :smile:

HHS81:
I do a merge by myself, as it works as it should thanks to wlbragg

gilbertohasnofb:
:+1:

gilbertohasnofb:
Hmm, I suddenly can't see the external shadow any longer, though the rain effect works as expected. Is it something with my installation or did something got lost in this PR?

wlbragg:
Mine is broke too.

gilbertohasnofb:
Ok, so something got broken, probably on this PR.

wlbragg:
No, I simply forgot to turn on shadows, mine is working.

gilbertohasnofb:
Oh, I see why now! The aircraft options dialog has an option to enable shadows by setting `/sim/rendering/shadow-volume` to `true`. There is no default value and when the sim starts that property is set to nothing, which FG understands as `false`. I think we should force it to `true`, given that the conditional inside `shadow.xml` will take care of renderings other than ALS. @HHS81 what do you think?

gilbertohasnofb:
@wlbragg I had posted before reading yours! Anyway, we should set the default to `true` IMO

HHS81:
What happens in default and rembrandt with set to true?

gilbertohasnofb:
I don't use Rembrandt, it does not really load in my system. As for the default renderer, it simply doesn't show any shadow but otherwise everything looks good.

HHS81:
I think we should/ could change it to true

gilbertohasnofb:
I wil take a look at it then.

gilbertohasnofb:
Indeed there won't be a problem with Rembrandt nor default renderer since the following conditional is used for the shadow object:

```
        <condition>
          <and>
            <not>
              <property>/sim/rendering/rembrandt/enabled</property>
            </not>
            <property>/sim/rendering/shaders/skydome</property>
            <property>/sim/rendering/shadow-volume</property>
          </and>
        </condition>
```

I am enabling the shadow and I will push in a sec

PR #131, opened 2016-07-26
Issue 124

gilbertohasnofb:
Closes https://github.com/HHS81/c182/issues/124

But before merging, I need your help, Heiko:

- the fuel caps on the top of both wings are not separate objects, but actually a single object with the wings. I need them to be individual objects to use the pick animation for the fuel dialogs
- there is a bug with the baggage door: only the external part of it is moving, not the internal. This means that the baggage door is not visible from behind when opened, and looks to be always closed from the interior view:

Here it looks normal:

From behind the door disapperas:

From inside you can see that the baggage door looks closed, but actually it's opened outside!


gilbertohasnofb:
Once those issues with the 3D model are fixed, I will finish implementing the fuel dialog and then we could merge this

HHS81:
@gilbertohasnofb 

> there is a bug with the baggage door: only the external part of it is moving, not the internal. This means that the baggage door is not visible from behind when opened, and looks to be always closed from the interior view:

I didn't check yet the PR, but I don't have any issues with the lugggage door in Master-branch.

gilbertohasnofb:
> I didn't check yet the PR, but I don't have any issues with the lugggage door in Master-branch.

You are right, it works in the `master` so it should be some mistake on my code. I will check it and push the fix in 15 minutes!

gilbertohasnofb:
Done. It was my mistake, I changed the capitalization of the Nasal function controlling the baggage door and it led to some bugs, sorry about that. So now it's only a matter of making those fuel tank caps into separate 3D objects so I can add the fuel dialog to them!

gilbertohasnofb:
:+1: just give me a second here now

HHS81:
@added clickspot to fuel tank cap.
I'm not happy with file structure for the baggage. Do we really need "./Interior/Baggage/baggage.xml" ?
I want to get rid of those nested file paths we can see so often, and seems to me to bring down loading and fps perfomance, and according to the wiki I'm right:

> Textures, transparency, .xml included sub-models: The things that kill frame rate in FlightGear. Try to avoid them.

[http://wiki.flightgear.org/Modeling_-_Getting_Started#Improving_models_for_best_FlightGear_performance](url)

gilbertohasnofb:
> added clickspot to fuel tank cap.

Great, I am taking a look on it right now, I will push the working dialogs very soon

>  Do we really need "./Interior/Baggage/baggage.xml" ?

What would you suggest? I don't mind changing it.

gilbertohasnofb:
Fuel dialogs done. Just let me know about the baggages path, so I can finish this PR and you can merge it then.

gilbertohasnofb:
@HHS81 Done. There is just one problem though, I think you forgot to make the hotspots on the fuel tanks invisible:

fg

After fixing that we can merge this

HHS81:
Making it transparent is one opteion. The another is to set the visible-tag to "true". I'm just doing it, then I'll merge it

gilbertohasnofb:
:+1:

gilbertohasnofb:
Perfect! So if there is nothing else to be done here, feel free to merge it :smile:

HHS81:
I meant "false". Anyway, merged!

PR #132, opened 2016-07-26
Issue 129

gilbertohasnofb:
Closes https://github.com/HHS81/c182/issues/129

I removed the weight of the copilot and passengers, and added a comment in the same line `<!-- <weight unit="LBS">180</weight> -->` so it should be quite easy to flip back and forth between them


gilbertohasnofb:
@HHS81 ping

PR #133, opened 2016-07-26
external shadow default as true

gilbertohasnofb:
Adds external shadows back


gilbertohasnofb:
Actually the external shadows default setting was already `true`, the problem was there were two blocks `<rendering>` inside of the block `<sim>`, and so FG created the second one as `/sim/rendering[1]/...`. Using a single block solves the issue.

gilbertohasnofb:
@HHS81 @wlbragg ping

PR #140, opened 2016-07-31
Radios to checkboxes in dialogs

### LINKED ISSUE #139: Aircraft Securing Dialog
@HHS81 Is there a reason why you mixed radios and check boxes in this dialog?

fgfs-screen-001

Given that all these objects have a binary state (either are there or are not), I think check boxes make much more sense. I think radios are normally use when you can select just one single option among N total options (e.g. the six variants of the c172p).

gilbertohasnofb:
Closes #139 

Changing radios into checkboxes on securing and aircraft dialogs.

Note that the ground-equipement dialog is fine, since the radio there is used to select under which wing the stairs will be put (only one option can be selected)

### DIFF
--- a/gui/dialogs/aircraft-dialog.xml
+++ b/gui/dialogs/aircraft-dialog.xml
@@ -64,7 +64,7 @@
 <group>
 	<layout>hbox</layout>
 	<padding>6</padding>
-	<radio>
+	<checkbox>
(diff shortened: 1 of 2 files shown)


gilbertohasnofb:
@HHS81 Thanks!

PR #145, opened 2016-08-01
Adding sounds to airplane

### LINKED ISSUE #127: add more sounds to the airplane
We can add more sounds to our plane. We can borrow some of them from the c172p, or I can come up with more of them (in the c172p, I did most of the work with the wav files, while @onox did most of the implementation in xml). The particularly nice ones are:

- rain and thunder sound
- parking break lever
- buttons and switches
- crash and damage sounds

gilbertohasnofb:
Closes #127 

Adds following sounds:
- adds external rain, rain on windshield and thunder sounds
- if doors or windows are opened, they affect volume of external sounds. Try opening and closing the doors while it's raining and with engine off. Then try the same with engine on
- all working switches have sound (master, avionics, toggles, gauges knobs, ELT)
- parking brake and trim wheel sound
- all working avionics have sounds (both KX-165, KT-76C, KR-87, KN-62A, GMA-340, KAP-140)


gilbertohasnofb:
:warning: :warning: :warning: Please don't merge yet! 

The heading indicator offset knob does not have sound yet because the property `/instrumentation/heading-indicator/offset-deg` keeps constantly changing, causing the sound to loop forever. Why is this property always changing, as well as a lot of others inside the path `/instrumentation/heading-indicator`?? Is this intentional? In the c172p I don't see this behaviour. So the sound for that only knob is commented out.

 @HHS81 If you could take a look at this, I'd really appreciate it. If you find the bug, please push the commit with the fix to this repo, so I can then add the sound for that knob and then we could merge this.

HHS81:
> Is this intentional? 
Well the c182s shows a bug here, the c172p maybe.

First: yes, the there is always a changing. That's why the HI has to be correct from time to time while flying. It is described in the dhc-6-thread in the forum. The question is, if the offset-deg is also changing with?

But: it all should stop, when engine stopped (vacuum driven) or when power off (elctrical gyro driven)
In the c172p it is vacuum driven, but I don't see the spin-property ever reaching 1. Anyway, I see a constant changing, but not the offset-deg.

The SenecaII has a electric driven one, it also constantly changes, but stops when power is off. 

The c182s should also have a vacuum driven one, but it seems to be a mix of electric and vacuum driven. I have to check this. 

The dhc6 uses two HI-models. One acts like in the c182s- constant moving of the offset-deg. The other one titled with "real" not. 

the dhc2 beaver has constant moving as well, though the model isn't used in the cokcoit (but specified in isntrumentation.xml)

So no idea, what is right.

gilbertohasnofb:
> So no idea, what is right. 

In that case, would you like to merge this and we think about it once we find an answer?

HHS81:
yes, the bug I discovered will be fixed later, as there seems more issues with the electrical system.

gilbertohasnofb:
Okay Heiko, thanks!

PR #155, opened 2017-08-25
Interior/magneto switch

hbeni:
Hi there,
i probably fixed HHS81/c182#42


hbeni:
(Just for the archive: i mislinked the PR so deleted this one.)

PR #156, opened 2017-08-25
Interior/magneto switch

hbeni:
Hi,
i problably fixed HHS81/c182#21


hbeni:
looks like this:
image

gilbertohasnofb:
I think this model looks way too long (also on the c172p project), see this photograph as reference:

hbeni:
I see it and try to fix it (the Problem lies in the interior model itself).

Thanks for the hint!


Am 2017-08-26 0:21, schrieb Gilberto Agostinho:

hbeni:
Gilberto notified me, that the model is too long, and indeed it is.
I will fix this in the interior model soon.

gilbertohasnofb:
@hbeni perhaps an easy fix would be to position the model further "inside" the cockpit, so that only a smaller portion is outside. But @HHS81 is the model specialist here ( :wink: ) so take my opinions with a grain of salt.

hbeni:
That was actually my plan :)

Am 2017-08-26 0:34, schrieb Gilberto Agostinho:

hbeni:
Gilberto notified me, that the model is too long, and indeed it is.

Looks better now:
image

HHS81:
Hi,

Thanks for the contribution, but this one unfortunately breaks the convention I use.

I'm not a big fan of loading every single models in the style by putting it in a new folder like here: Interior/Panels/Magnetoswitch...
Exceptions: for reuse; being not explicitly part of the aircraft itself (like carts, chocks etc...) or special animation (different panel layouts like EC 135 P2)

The reasons are: 
Perfomance: the more submodels/.xml-models loaded, the more impact on framerate. So I try to keep this number low

Rendering: the magnetoswitch here isn't affected by the ALS-shadows/ lights. For this it have to be part of the interior.ac-model and animations should be inside interior.xml.

Maybe I will manage it today to adapt your contribution to this convention.

P.S.: If I would be Gilberto, I would also claim that the [cessna keys looks different](http://www.ebay.ie/itm/NEW-CESSNA-Aircraft-OEM-Factory-Key-Blank-Uncut-Part-152-172-182-206-210-310-/112393332201?hash=item1a2b2a25e9:g:JJ4AAOSw9uFW9JoH)  😄

gilbertohasnofb:
> P.S.: If I would be Gilberto, I would also claim that the cessna keys looks different :smile:

Oh, nice catch! I will post this on the c172p repository :smile:

hbeni:
> Exceptions: for reuse; instruments (as it is often easier to animate but then it suffers from the problem described below); being not explicitly part of the aircraft itself (like carts, chocks etc...) or special animation (different panel layouts like EC 135 P2)
>
>The reasons are:
Perfomance: the more submodels/.xml-models loaded, the more impact on framerate. So I try to keep this number low

Hello,
i totally understand this, i was not aware and thought that the current structure may be adapted to the c172 style because it  eases maintenance of the code (in my opinion).
Is the performance draft really that huge? And when does it occur (i would not expect it to happen during runtime but only loading time since it is in-memory after the load).


If you would please tell me what exactly needs to be done, i would gladly help you out with my time and adapt it to the current structure.
(is it: move XML code to -set.xml and integrate key model in the interior-model?)

HHS81:
> i totally understand this, i was not aware and thought that the current structure may be adapted to the c172 style because it eases maintenance of the code (in my opinion).
Is the performance draft really that huge? And when does it occur (i would not expect it to happen during runtime but only loading time since it is in-memory after the load).

For good framerates GPUs prefers ro tun few batches as possible. Means less objects/meshes and textures as possible. The size of those objects and textures are not that important like their numbers. That's why it is recommended to combine all parts of a model in one single mesh/object, the same for the textures[1]. I made tests where I loaded 1000 single mesh cubes, and then a combined mesh of 1000 cubes. The later one gave me a much higher framerate.

When you split up everything into submodels, you will have more objects and with a higher impact. 
I once made another test on the scenery: I placed 1000 simple houses in the scenery. First just the .ac-model. Then  .xml lodading the .ac-model.
Actually the plain .ac-model gave me a much better fps than the per .xml-loaded building. 
Another author noticed it as well and mentioned in the wiki [2]. I guess the reason is, that .xmls are handled in a different way than .ac-models, since those .xmls must be read continuously.

Another example is [Gary R. "Buckaroo"Neely's Veloycity XL.](http://www.buckarooshangar.com/flightgear/index.html).
A very outstanding model due to its details and its graphic. Compare it with the Velocity by Helijah from FGAddon and take a closer look on the framerates. The Velocity by helijah is less detailed, though framerates are not better - indeed it is the opposite, they are worse! Now compare the file structure and the way Gary modelled the aircraft.

When I create aircraft, I think of common users. They are not interested in maintenance. They are interested in flying - with the best fps they can get. 

[[1]](http://wiki.flightgear.org/Modelling_guidelines)
[[2]](http://wiki.flightgear.org/Modeling_-_Getting_Started#Improving_models_for_best_FlightGear_performance)

HHS81:
P.S.: I haven't combined yet all meshed - an open issue, will be closed when we are sure there is nothing left to do.

HHS81:
> Hi,
i updated the branch to master and integrated the key model and its animation in the base files - i hope its ok like this or did i miss something?

Not quite yet ;-) 
Nitpicking:
I can see from the colors that the key-models are not affected by the lightmap:
fgfs-screen-005
So I made some smaller changes:
"start"-object uvmapped to PanelTexturing-MainPanel.png.
Removed the unused material-animation from mangneto-switch in interior.xml

I agree that we can add the correct key-model later, and add it to the right uvmap, so it is affected by the lightmap.


It would be all easier, if we wouldn't need lightmaps. But since Rembrandt seems deprecated, it is the only way to get something nice looking interior-lights in ALS. So everything must be uvmapped to a lightmap, and this in a intelligent way, since we only have 4 channels.

hbeni:
Hello and thank you for merging.
Sorry that i did left over so much work for you, but i don't have enough knowledge of 3D-modelling yet.
I think the main advance now is, that one can operate the magnetos now.

HHS81:
Thanks for your contribution!
It wasn't that much work. And yes, operating the magnetos is another great step!

hbeni:
Just for the archives: What is missing here is just the correct key model.

Issue #157, opened 2017-08-25
Support custom immatriculation


hbeni:
It would be nice to have a custom immatriculation like in the c172

PR #158, opened 2017-08-25
Livery/custom registration 1

hbeni:
Hi i recently added a real life livery (D-ERLI) with support for custom registration.
Fixes HHS81/c182s#157


hbeni:
Image of D-ERLI
Real life D-ERLI

d-ebhx
Livery (i skipped the advertising and country flag)

HHS81:
Hi,

Thanks for this contribution!
The "D-ERLI" is actual a Cessna 182 **T** model, which is a successor of the S-Model we simulate here. You can notice the difference by the small additional horizontal stab on the fin and the different shape of the air intake at the front of the cowling.

But since the original registration isn't shown here, I don't mind a "wrong" livery. With the  "D-ELFP" we have already another "wrong" livery.

I should have modeled the T-model from the very beginning, more interesting liveries ... *sigh*

hbeni:
Thank you, i was fully aware of this however wanted custom callsign support. And since i know ERLI from real life, and like the painting i thought it would be a good thing to use it.
Besides that, the wheel cover also is a little different at the T.

HHS81:
It is a very nice livery, well done!  I really like it!
And thanks for the custom callsign suport, I think this will attract more people to this little GA!

Yep, the wheel fairings are different too. In the cabin there is a difference on the domelights and cautionlights are placed different (for the non-G1000 version). Maybe I will add this version some time, but I hope for a good G1000-model.

gilbertohasnofb:
@hbeni Just a couple of issues and ideas concerning this livery:

- perhaps you could redo the nose and tail as there are some strange places:



- maybe the blue colour could be a bit less saturated, but perhaps that's only my taste

- the callsign is actually "floating" and not touching the sides of the plane:

Issue #151, opened 2016-08-10
Add sound to fuel pump


gilbertohasnofb:
When toggled on, the fuel pump has a very characteristic sound, listen to 1:20 in the video below: https://www.youtube.com/watch?v=-yVMskFQZsw

gilbertohasnofb:
@HHS81 at which moments should one the fuel pump, only for a few seconds in the startup?

HHS81:
Usually always,  except the engine is still warm(<20min). I don't see much difference between cold and hot weather. I will probably try on the weekend to implement a more realistic fuel system which will be important for the realistic engine start.

HHS81:
Where we get such sound?

gilbertohasnofb:
> Where we get such sound? 

I can try to recreate it either by synthesizing it or by using some other similar public domain sounds as source. I will give it a try in the next days.

gilbertohasnofb:
@HHS81 sorry I still didn't have the time to tackle this, but I hope to do it soon. Also I wanted to ask you if our avionic switch should also create this high pitched constant sound whenever it's in the ON position (see 2:21 in the video below):

https://www.youtube.com/watch?v=LeuGAxTA4fg

And would you know if this sound would also be present in the 172P by any chance?

HHS81:
Sorry for the late answer, but I'm busy too and needed a large pause of flightgear. I currently re-arrange my mini-appartmement, so I get more room for a sofa and larger dining place. 

About your request: Unfortunately the video isn't available anymore. 

But there are two sounds which you will hear when avionics switch is enabled:
- the cooling fan of the avionics
- the gyro

And yes, those sounds will pre present on the c172p, I'm sure.

HHS81:
I guess I can start with work on in about one or two weeks, my computer crashed, I had to reset it, and I could not recover yet all of my programs. All a bit too much right now....

gilbertohasnofb:
@HHS81 No problems Heiko, I also have been extremely busy lately. But today I finally found the time to synthesize that sound sound after analysing several recordings in those videos, and it's sounding quite decent if you'd ask me. Please check the file, it's in the branch `Issue-151` and the file is called `Sounds/avionics.wav`, and let me know what you think.

gilbertohasnofb:
>  I currently re-arrange my mini-appartmement, so I get more room for a sofa and larger dining place. 

And by the way, good luck with the apartment!

[...]

PR #162, opened 2017-09-10
Issue 151

HHS81:
This PR add fuel pump sound. When engaged you hear the looped sound, when switched off pump-end.wav will sound.
Fix Issue #151


gilbertohasnofb:
@HHS81 Ok, so I tested this and here are my comments:

- the fuel pump sound implementation works great, thanks!
- when I load the sim, I can already hear the gyro sound, likely because the avionics and master switches are all on (by the way, I can't see any option to start the aircraft in cold and dark conditions, we should add one). After turning off those switches, the gyro sound takes ages to die off. I assume this is a feature, so can you just confirm this is intentional? If so, all good.
- at low throttle, the gyro sound is easily cutting through the sound of the engine. I don't think the issue is with the gyro sound but rather with the engine sound: perhaps the volume at low throttle should not be that soft.

HHS81:
@gilbertohasnofb 
> After turning off those switches, the gyro sound takes ages to die off. I assume this is a feature, so can you just confirm this is intentional? If so, all good.

Yes, the spinning of the gyro is hardcoded and seems realistic to me. So yes, it needs a while until the spin has stopped.

> at low throttle, the gyro sound is easily cutting through the sound of the engine. I don't think the issue is with the gyro sound but rather with the engine sound: perhaps the volume at low throttle should not be that soft.

I'm not sure which of the instruments makes noise in reality. According [this source](http://www.pprune.org/private-flying/426458-turn-coordinator-gyro.html) it does. So I aligned to every instruments, which has <spin>-tag the gyro sound. 
I can set the volume of the gyro down, and increase engine volume with low throttle

gilbertohasnofb:
> I can set the volume of the gyro down, and increase engine volume with low throttle

Perhaps that would be a good solution, and then we merge this. Everything else is looking good then!

I will also open a new issue for the cold and dark option.

HHS81:
@gilbertohasnofb 
O.k., only the Turn-indicator uses an electrical gyro according POH, So I removed the other ones from the sound.xml.

The interesting finding: our FGFS heading-indicator is electrically driven. But it should be vacuum driven according POH. I wonder where to change it.

Anyway, please test it!

gilbertohasnofb:
> O.k., only the Turn-indicator uses an electrical gyro according POH, So I removed the other ones from the sound.xml.

Sounds good to me, I will test now.

> The interesting finding: our FGFS heading-indicator is electrically driven. But it should be vacuum driven according POH. I wonder where to change it.

Will you open a new issue about it so we don't forget it? I did open one for the startup state options.

hbeni:
this is already fixed #151

Issue #107, opened 2016-07-17
implement complex startup procedure


gilbertohasnofb:
This is what the checklist for our model says about starting the airplane:

> STARTING ENGINE (With Battery)
> 1. Throttle – OPENING 1⁄4 INCH
> 2. Propeller – HIGH RPM
> 3. Mixture – IDLE CUT OFF
> 4. Propeller Area – CLEAR
> 5. Master Switch – ON
> 6. Auxiliary Fuel Pump Switch – ON
> 7. Mixture – ADVANCE to obtain 3 to 4 seconds fuel flow, then return to IDLE CUT OFF position
> 
> NOTE
> If engine is warm, omit priming procedure of step 7 above
> 1. Ignition Switch – START (release when engine starts)
> 2. Mixture – ADVANCE smoothly to RICH when engine fires
> 
> NOTE
> If engine floods, turn off auxiliary fuel pump, place mixture In idle cutoff, open throttle 1⁄2 to full, and crank engine. When engine fires, advance mixture to full rich and retard throttle promptly
> 1. Oil Pressure – CHECK
> 2. Auxiliary Fuel Pump – OFF
> 3. Navigation Lights and Flashing Beacon – ON as required
> 4. Avionics Power Switch – ON
> 5. Radios – ON
> 6. Flaps – RETRACT

Currently, our airplane is only checks that the mixture is high enough to start the engine, even if the master switch is off. The throttle lever gets ignored and the engine starts regardless if it is at 0% or 100% or anywhere in between. The propeller pitch lever also gets ignored. The engine also starts even if the Auxiliary Fuel Pump Switch if off. Finally, the complex mixture lever procedure described above could also be implemented.

gilbertohasnofb:
Personally, I consider this to be a high priority issue given that our aim is at a realistic plane and not a "press S to fly" kind of plane.

HHS81:
I consider this as an after-FGAddon Release point because:
1.) This needs a complete simulation of the fuel system which is very different from c172p as the newer Cessnas uses fuel-injected engines.There is no example yet in FGFS.
2.) This also needs changes to engine and prop files, since the current numbers at idle aren not correct which makes the engines start immediately and showing too low mp. Ich already

gilbertohasnofb:
> I consider this as an after-FGAddon Release

That's why I added the milestone FGAddon Release **2** :wink:

gilbertohasnofb:
[...]

Issue #154, opened 2016-11-19
Wing strut not connecting properly with the wings


gilbertohasnofb:
Wing strut not connecting properly with the wings, see:

![fgfs-screen-001](https://cloud.githubusercontent.com/assets/5700795/20459028/c91eecb2-aeb4-11e6-91da-a57342c49fd2.png)

Notice that there are small holes in between the strut and the wing

hbeni:
Hi, i fixed this in the fuelSelector branch with commit 7226c32

Issue #123, opened 2016-07-25
Implement autostart function


gilbertohasnofb:
Right now starting the engine is trivial, but once we have a realistic startup simulation then we should add a autostart function (which should also take care of oil level, securing elements and static objects, fuel contamination, fuel quantity, etc.)

gilbertohasnofb:
Once we have a more realistic startup simulation, I can implement this myself.

hbeni:
I think this could be implemented easily via the automatic checklist execution once #121 is ready

hbeni:
fixed with  6ba98ea

Issue #109, opened 2016-07-17
Save states between flights


gilbertohasnofb:
One of my favourite features of the c172p is that when you activate the save state toggle in the aircraft menu, every toggle and switch and knob and lever remain in the same position next time the sim is loaded. It brings a great deal of realism to the sim, and I would say we should consider doing the same for ours.

HHS81:
yep, for sure. The same about fuel state, battery state etc.....

hbeni:
d7ba4fc adds many options to the save-state of the aircraft; i think i got the most items already.
What is not working right now is the state of the avionics busses and battery switches  - are they overridden in some nasal script?

HHS81:
They are set in `c182s-electrical.nas`

hbeni:
Should be fixed in PR #167 (https://github.com/HHS81/c182s/pull/167#issuecomment-331739868) already :)

hbeni:
Open from the pull request is the ground equipment/aircraft securing

hbeni:
In my experiments, aircraft securing now also works. What still does not work is ground equipment like the ladder etc.

PR #167, opened 2017-09-18
Complex startup procedure

hbeni:
Implemented complex startup procedure based on the interior/fuelSelector branch.
-> implemented #107 

The only thing not yet correctly working is starting with 0 mixture setting.
see the checklist, it says:
- engage starter
- once engine fires, enrich mixture smoothly to full.
(see also linked video at the start)

=> So as long there is fuel in the manifold tank and the engine is primed (due to explicitely priming OR warm temperature), the engine should start. Currently it needs a certain mixture threshold.
Where (in which file) is that threshold defined?
It would be good if the engine would just suck from the manifold and apply the mixture setting at a short time after the engine started?

The procedure seems to be the same as when in the engine-flooded state, except for the need to have throttle at least 50% open.


hbeni:
Where (in which file) is the mixture >50% threshold to start the engine defined currently?

HHS81:
It is hardcoded by jsbsim. But there could be a workaround, at least when I remember. Need some time to find.

HHS81:
Btw. While priming the amount of throttle also affects the fuel flow. With full throttle you would have full fuel flow. According this video https://youtu.be/g3pbezIuimw

HHS81:
@hbeni  there you go: https://github.com/jentron/c182/commit/325a18cbad177d242c8d06fd792acad6dfcf3983#diff-4e1e4f6e6c5c7dd8181adda4fdcef8f9

The last commit on c182s.xml does have an example of a mixture Code which might could do the thing. Beware that the real engine will die soon as mixture is = Zero

hbeni:
@HHS81 we are nearly there, the POH operations are now fully supported. :) (yeah! that was hard work)

I made the priming ammount depending on outside temperature, is this correct? (rather then oil temp?)

Still unsolved is also that the fuel flow from pump when priming should rise with pressure rather thatn the mixture lever (i.e. the fuel flow lagging behind the mixture a bit and being smoother, not directly linked to the mixture lever). This can be solved later on probably tough and may not be a blocker for merging (what do you think?).

hbeni:
@HHS81 @gilbertohasnofb I did more work today and i think it looks very promising.
Most notably, the priming depends on oil temperature (which depends on outside temperature) now.
I also added code so the fuel transfer between tanks honor gravity forces now.

The last things open at the moment:
- [ ] Oil-Temp in the engine seem to always initialize at about 50°F. This currently yields slightly wrong priming need. Is this hardcoded in jsbsim, or can we change it? (where?)
- [ ] The fuel flow gauge when priming should be more smooth and resemble a bell courve. Currently it is directly tied to the mixture contol flow setting which gives "instant" flow.

Please feel free to test :)

HHS81:
O.k. tested.
I needed several attempts to start the engine - I always overprimed. 🤣 
It was my fault, since I didn't exactly follow the checklist (more than 1/inch throttle, too much mixture, etc....).
Like the real thing it can be tricky. I'm not sure if the real thing is tricky like that.
So we indeed need something like a magic button for beginners. (Maybe like the EC 130 B4 helicopter which enables autostart, when a certain amount of successful starts AND successful landings had been achieved. )

HHS81:
So shortly said: Really great work, I love it! :smile:

The implemenation of a fuel injected engine seemed much more complex, as I ever thought. You  ight want to write to the [JSBSim mailinglist](https://sourceforge.net/p/jsbsim/mailman/) and show your work there. It could be interesting for the devs there.

hbeni:
You made a good point with the beginners button.
I always hated the implementation at the helicopter tough.
It is much more user friendly to make a simple checkbox like in the c172p. As this is also easier to implement, i think i integrate this in the airplane options dialogue and make it "on" by default.

hbeni:
Btw im still fiddling with the tank level maintenance which confuses the autopilot in some situatins (slow speed).

HHS81:
Yes, the checkbox seems like a good idea to me

gilbertohasnofb:
> I always hated the implementation at the helicopter tough.
> It is much more user friendly to make a simple checkbox like in the c172p. As this is also easier to implement, i think i integrate this in the airplane options dialogue and make it "on" by default.

:+1: 

On the c172p, we have both a toggle for complex/simpler engine operations as well as a single menu item Autostart. I think that's simple enough and solves all cases (beginners, advanced pilots, lazy ones).

hbeni:
The latest commits (fd7ba4fc, a88bd0, 78b8827) should fully implement #109 :)

Also adjusted the tank-transfer rates again. It should be ok now i think.

@HHS81 Open is still the better display of the fuel flow gauge - or do you think it is ok already?

Next step is to create a nasal script to support autostart. I think an easy approach would be to
- get the current state of complex engine procedures for later restore
- disable it if set
- set mix and throttle for startup
- set magentos and activate starter unil engine running
- release starter
- restore complex engines procedures setting

Do you think this would work?
@gilbertohasnofb can you help me out with this? I would head over to the checklists then.

HHS81:
@hbeni  

> Open is still the better display of the fuel flow gauge - or do you think it is ok already?

What's wrong with the fuel flow gauge?

hbeni:
@HHS81 
> Cool! Would be nice, if the ladder, fuel tank trailer, external power, chocks and safety cones could be state saved as well.

Thanks for the hint, i will add that!

> What's wrong with the fuel flow gauge?
Currently, when priming, it raises stepwise when pushing in the mixture control, but i think it should be a little more smooth. This is no big issue and just cosmetics.

I'm on a business trip during the week and cannot work on this more at least until the weekend.
For the autostart suff i currently still lack the nasal knowledge (and i suppose we need nasal here). it would be cool if somebody could dig into this while i'm away. :)

hbeni:
Well, i think we should be there now!
- Autostart is working now (it disables complex engine procedures, so battery state, temperature, priming etc does not matter at all and engine should start reliably)
- Complex engine procedures are implemented; a basis for fuel contamination and oil management is there
- Battery state is saved correctly now

hbeni:
@HHS81 @gilbertohasnofb 
From my point of view, the branch should be ready to be merged and will fix several open issues (#52, #107; and on the fly #154, #123 and #109 (95%), ).

hbeni:
Please test this and after the merge i would like to have a try on the checklists :)

HHS81:
@hbeni 
I allowed me to remove the static objects from the save state, since this isn't unfortunately not working.
I also removed the lines with the out-ceommented carburetor for clean up.

HHS81:
@hbeni 
The autostart is working great. 
I think we need also an autoshutdown. On most FG aircraft it is common that the Autostart-button also works as Autoshutdown. I tried something by adding a line, and changing the message, and it works. Do you want it?
When the aircraft is started, mixture should be full rich for takeoff. Do you want me to change this?

gilbertohasnofb:
> On most FG aircraft it is common that the Autostart-button also works as Autoshutdown.

Really? What other aircrafts have this feature? Am I right to think the c172p is not one of them? Anyway, I am always against any sort of cryptic features, why on Earth would something called Auto __start__ be used for shutdown? Isn't it better to have a second option in the menu called Autoshutdown, for the sake of making all of this completely obvious for the user? For instance, I myself am using FG for years now and have been actively participating in many projects and yet I have never heard of such thing before, which perhaps is a tell that the idea is a bit too obscure.

HHS81:
@gilbertohasnofb 
Most of the airliners do.
You can simply label the button as "_Auto Startup/Shutdown_". I think Users are (mostly) intelligent enough to understand how it works

gilbertohasnofb:
> Most of the airliners do.

So that's the reason I never saw it, I rarely touch the airliners.

> You can simply label the button as "Auto Startup/Shutdown". 

This is a good idea, it would make things very clear. My issue was with calling it Auto Start only and then expecting the users to know about a secondary function.

>  I think Users are (mostly) intelligent enough to understand how it works

Maybe I am not one of those :stuck_out_tongue_winking_eye:

hbeni:
I think an autoshutdown would be a nice feature for the beginners, however wouldn't it in essence just be setting mixture to 0%? (or are we talking about a shutdown-to-cold state where also battery, avionics etc are turned off like described in the checklists?)

I agree that the shutdown function should at least be visible from the start menu entry.
@HHS81 go ahead and push your changes directly to the branch :)

HHS81:
@hbeni 

> I think an autoshutdown would be a nice feature for the beginners, however wouldn't it in essence just be setting mixture to 0%? (or are we talking about a shutdown-to-cold state where also battery, avionics etc are turned off like described in the checklists?)

Depends on what we want.
- just setting mixture to 0% is adding just one line.
- shutdown-to-cold state needs more, but would be like the now implemented autostart more realistic

hbeni:
Have you already implemented some local-to-your branch code?
If not i would implement it as following:
- combined autostart/autoshutdown and labeled as such in the menu
- shutdown to cold state according to checklists

HHS81:
@hbeni 

> If not i would implement it as following:

    combined autostart/autoshutdown and labeled as such in the menu
    shutdown to cold state according to checklists

Each as an own button?
Or  combined autostart/autoshutdown = autostart ccording to checklists and shutdown to cold state according to checklists

hbeni:
>  combined autostart/autoshutdown = autostart ccording to checklists and shutdown to cold state according to checklists

this

HHS81:
o.k. I will implement it like that

hbeni:
cool!

HHS81:
O.k., I added also Ground securing (not part of the official checklist); should I remove it?

hbeni:
Good question. I think we can leave that in; the target audience of shutdown is probably novice pilots and it is helpful to them to see the plane tied down.
However, once checklists are in place, they also could just click trough the checklist items...

HHS81:
O.k, I was too fast here... 🤣 
I enable ground securing again 😄

HHS81:
Done! 
If everyone agree I will merge it - our do you @hbeni want to have the pleasure?
Since it is mostly your fantastic work, I think you do it! 😄

hbeni:
Do you mind that i close the associated tickes after the merge too :)?

And do you mind if i add myself to the authors list ;) ?

gilbertohasnofb:
> Do you mind that i close the associated tickes after the merge too :)?

Just so you know: if you list the tickets in the first message when opening a PR, they will all be automatically closed once the PR is merged.

HHS81:
@hbeni 
If someone not already did, you have to add your name of course ;-)

hbeni:
Hooray, its done :)

gilbertohasnofb:
:clap: :tada: Thanks a lot for all the hard work!

hbeni:
@HHS81 Do you want merged branches to be deleted?
@gilbertohasnofb I leaned alot and it was fun (most of the time ;) )

HHS81:
@hbeni Great! Many tnanks for your great, but hard work! 🍾

HHS81:
@hbeni I think we can merge it!

HHS81:
Now it is my task to adjust idle-rpm and with that power and fuelflow!

hbeni:
Merge is in and tested. All works fine it seems 👍

Issue #121, opened 2016-07-25
Add checklists (clickable from the cockpit)


gilbertohasnofb:
- [x] Add checklists
- [x] Make them accessible from the cockpit

gilbertohasnofb:
We can also add a 3D model of the checklist in the pocket on the left side of the pilot's pedals and open the checklists once the user clicks there (similar to the c172p).

hbeni:
Startet workin in a private branch. I will push this as soon as i have a somewhat stable structure.

hbeni:
Engine start checklist is pushed: https://github.com/HHS81/c182s/tree/Checklists
Please have a look if i'm on the right track here.

hbeni:
Checklists in place, but still open:
- [ ] 3D Pocket model for easier checklist access (like in c172)

hbeni:
@gilbertohasnofb look at this link: https://youtu.be/TI1zpg0zGug#t=07m06s how would one model that?
Just a simple shaped 2D plane and the book separately as simple block? Or both as 3D shape?
Maybe i can convince blender to do what i want some day... in 2018

gilbertohasnofb:
> how would one model that? Just a simple shaped 2D plane and the book separately as simple block?

@hbeni I would say that it depends on how detailed you want it to look. We could simply steal the model from c172p and change the texture of the manual in case the 182S's POH has a different cover than the 172P's. See:

[img]https://s4.postimg.org/eoc55wl4d/fgfs-screen-003.png[/img]

Would you have something else in mind?

hbeni:
Ok, again i failed with blender.
i'm giving up on this now. :(

hbeni:
For reference:
![image](https://user-images.githubusercontent.com/13608602/32138562-e44bb8ba-bc34-11e7-926c-4d7eb36e25f6.png)

HHS81:
In the making - found some space in the cabin aka uvmap.....

gilbertohasnofb:
> In the making - found some space in the cabin aka uvmap.....

That's great news! I can't believe you could squeeze anything else there 😆

HHS81:
@gilbertohasnofb 
Well, it isn't really high resolution... 😞 
![fgfs-screen-005](https://user-images.githubusercontent.com/12649375/35779289-e0709998-09ca-11e8-85af-0a5b3c049567.jpg)
But I think it is good enough.....

gilbertohasnofb:
[...]

PR #169, opened 2017-10-17
Checklists

hbeni:
Implemented all important checklists, fixing #121 

@HHS81 @gilbertohasnofb Ready for your testing i think :)

Still open:
- [ ] 3D Pocket model for easier checklist access (like in c172)


hbeni:
gilbertohasnofb said in the issue:
> We can also add a 3D model of the checklist in the  pocket on the left side of the pilot's pedals and
> open the checklists  once the user clicks there (similar to the c172p).

That sounds like a great idea. Should we just port over the pocket model and integrate it in the interior model?

gilbertohasnofb:
> That sounds like a great idea. Should we just port over the pocket model and integrate it in the interior model? I just dont't know how this works with the texturing. 

Yes, I think that could work. But, unlike the c172p, we try to have as few .ac and texture files as possible as that really improves performance. Perhaps @HHS81 can integrate that checklist model into the interior model, and if he would UV-map it I would gladly create a texture for it.

PR #170, opened 2017-10-18
Correct priming fuel flow gauge

hbeni:
I found a bug with the new complex engine operations for priming oncerning the fuel flow gauge.
According to the video https://www.youtube.com/watch?v=-yVMskFQZsw the fuel flow needs about 3-5 seconds to become stable, at which point the pilot should put the mixture back into idle cut-off.
Until now i had the understanding, that after the fuel flow is stable, the pilot should wait 3-5 seconds, but this is way too much and causes flooding.

I fixed this in this and now the behavior is correct in two ways:
- the fuel flow gauge now follows a courve when priming. The top point is about where the engine is primed correctly, like in the video.
- the fuel flow (when priming) is feed through a kinematic, which makes the needle move more realistic and appealing to the eye

Oh yeah, and please note that this system is already sensitive to oil temperature, so once this initializes to the outside-air-temperature when starting, priming time varies with the weather.

@HHS81 @gilbertohasnofb if you don't mind, i would merge this, ok?


HHS81:
@hbeni  sure, it is o.k. to merge it! :-)

hbeni:
Great, i will do it now.

I just wanted to ask as you are the maintainer and im not sure of the merge 
policy in place :)



Am 2017-10-21 21:39, schrieb HHS81:

hbeni:
Oops, that merged the checkists too, because i overlooked, that i forked off 
the checklist branch when correcting this.
I was ready with the checklists however, so no harm damage have been 
inflicted by this.

Am 2017-10-21 21:39, schrieb HHS81:

Issue #149, opened 2016-08-01
Problems with the radio stack


gilbertohasnofb:
Here is a list of a lot of issues, bugs and suggestions for improvements concerning the radio stack:
- [x] the ADF knob has a very good working knob, in which we can intuitively scroll up and down using the mouse wheel only, regardless if we are using the inner or outer knobs. On the other hand, the COMs, NAVs and KAP140 require using the `Shift` key for the outer knob. I would suggest we standardize this and use the ADF standard.
- [x] the model of the outer knobs of the COMs and NAVs do not rotate
- [x] switching between stand by frequencies on COMs and NAVs should not move the frequency knobs, I believe
- [x] the models of both inner and outer knobs of the KAP140 do not rotate
- [x] the models of both inner and outer knobs of the DME (KN62A)do not rotate
- [x] DME (KN62A) and autopilot (KT76C): scrolling with the mouse wheel does not work on their knobs
- [x] DME (KN62A) toggles have anti-intuitive behaviour: the upper one moves to the right with left-click and to the left with right-clicking, but the bottom one flips back and forth using just left-click while right-clicking on it does nothing. An ideal solution would be using only left clicking on all of them.
- [x] ~~~GMA340 rotating knobs not implemented~~~ opened #174 to keep track of this

hbeni:
recently fixed by https://github.com/HHS81/c182s/pull/180

Issue #128, opened 2016-07-25
Improvements to the Radio Stack


gilbertohasnofb:
Recently, @wlbragg and I did a lot of improvements in the radio stack of the c172p. We could check if those improvements are not fit for the c182 as well.

hbeni:
fixed recently

PR #175, opened 2017-10-25
Various cockpit improvements

hbeni:
This This evening i improved the cockpit alot.
#128 is fixed, the radio stack now operates like in the c172 with MUCH more userfriendlyness.
#149 is mostly fixed, however the Garmin device is still not operable. This should be tackled in a separate issue that i opened just now.

Please test :)


hbeni:
@gilbertohasnofb It seems that the KN62 ADF radio is too dark; where/how can i adjust the baseline illumination?
image

HHS81:
@hbeni change the material settings in the .ac-model. (via text editing)

hbeni:
Ok, thanks, fixed this in commit 870da5e

hbeni:
Ok, i think this is ready to merge now.
@HHS81 If you give your OK, i will merge this into master.

Issue #53, opened 2016-01-31
Make Davtron 803 working


HHS81:
- [x] make Davtron 803 working
- [x] add texture for horizontal lcd line and display selected mode

hbeni:
Hi @HHS81 this evening i did have a try with blender. I failed horribly :)
I need your modeling skills urgently; the davtron model just needs some faces to display the numbers later on (like done in the immat model). I will then do the rest.

Please, i need a working 3D-model with the following named objects (2D-planes with proper material and texture linked).
I already supplied the numbers textures in the branch Davtron803. For correct scale and place, please see [google images](https://www.google.de/search?q=davtron+m803&safe=off&client=firefox-b-ab&dcr=0&source=lnms&tbm=isch&sa=X&ved=0ahUKEwi8_cPqlYDXAhXLJMAKHQMGCe8Q_AUICygC&biw=1352&bih=816) (note i did not find any picture showing negative degrees in the top LCD window; please just put it far to the left side for now):

```
    <!-- top LCD elements -->
    <object-name>lcd.top.minus</object-name>
    <object-name>lcd.top.1</object-name>
    <object-name>lcd.top.2</object-name>
    <object-name>lcd.top.dot</object-name>
    <object-name>lcd.top.3</object-name>
    <object-name>lcd.top.mode</object-name> -->
    
    <!-- LCD divider -->
    <object-name>lcd.line</object-name>
    
    <!-- bottom LCD elements -->
    <object-name>lcd.bot.mode</object-name>
    <object-name>lcd.bot.1</object-name>
    <object-name>lcd.bot.2</object-name>
    <object-name>lcd.bot.dots</object-name>
    <object-name>lcd.bot.3</object-name>
    <object-name>lcd.bot.4</object-name>
```

gilbertohasnofb:
If one of you guys create this model, I would be happy to help with textures if anyone needs it.

hbeni:
The model is already there, what is missing is just the 2D-Planes to show the LCD-display content.

hbeni:
In the instrument model folder, i added some LCD text band textures already:
 - lcd1.png shows numbers from 0 to 9
 - lcd2.png contains the mode chars for the temp modes (1/E: Voltage, 2/F: OAT in degF; 3/C OAT in degC) as well as special chars (4/":" for the divisor of the time; 5/"." for the dot in degF/degC; 6/"-" as minus sign in case of negative degC).

[...]

PR #177, opened 2017-10-28
Davtron803

hbeni:
Implement Davtron 803
fixing #53


hbeni:
@HHS81 I think this is ready for merge now.

PR #179, opened 2017-10-28
Added Clickspots to switches (fixed #173)

### LINKED ISSUE #173: Make switches more easily to click
currently the switches are hard to hit, especially when using head tracking:

image

We should add some clickspots there. I already looked into the model to see if the switch bases are adressable, however they are unfortunately part of the panel object, so we need to add some clickspots in the model and cant fix this with XML only.

hbeni:


hbeni:
Hi @HHS81 i want to merge the current state of the branch because on of the two issues is already fixed.
The branch could be deleted afterwards as i'm not able to add the pocket model. If you have time sometime, it would surely be better to fork off a new branch from the then-master.

PR #184, opened 2017-11-14
Fuel contamination

### LINKED ISSUE #126: implement fuel contamination
Implement fuel contamination, create a dialog for checking fuel and create a hotspot below the wings  (see c172p for an implementation of this idea)

hbeni:
Implemented fuel contamination in case "complex engine procedures and failures" is on (fix #126)


hbeni:
I did not find bugs anymore, so i think this is ready to merge. If you don't mind i will merge it soon.

@HHS81 I also ported the surfaces preflight checks from the c172, hope that was okay.

hbeni:
Also fixed missing button lighting of two avionics

HHS81:
It is very noticeable that the infrastructure of the c172p had been used here - the `.../active-engine/...` comes from the fact that they used two different engines which can be changed at runtime. But I don't think it is a showstopper here.

hbeni:
@HHS81 no problem here, if you want i can move over the properties to engine[0] :)
That shouldn't be a big deal and if it makes the code more clean, would be a good thing to do IMHO.

PR #185, opened 2017-11-20
FuelSystem: Corrected negative G-forces cutoff to real zero gravity

hbeni:
I think for a top fuel tank system the zero-g cutoff must occur relatively near to 0 g. -0.5g is way too much it seems.


HHS81:
@hbeni 
Great! Many thanks!

Issue #125, opened 2016-07-25
Implement oil management


gilbertohasnofb:
Implement oil management, create a dialog for checking oil level and create a hotspot around the place the oil tank cap in the aircraft nose (see c172p for an implementation of this idea)

gilbertohasnofb:
I wanted to start working on this, but here are a couple of questions before:
- where exactly is the oil tank on the 182S? Is this the cap?:
![](http://cdn2.planeandpilotmag.com/stories/2008/june/ntsb-IMG_1790.jpg)
- if it is, I think we should add it to the liveries, or better yet to a heightmap like we did for the c172p
- once we know where the oil tank cap is, to implement this, I'd need help from someone to create a hotspot just like @wlbragg did for the c172p

gilbertohasnofb:
Info for myself: oil tank can take up to 9 quarts, critical value is 4 quarts

HHS81:
The oil cap is on the right fuselage in front of the door. It is drawn on the texture and included by the normalmap I think.

I can add a hotspot but not today. 

Please consider a use of different types of oil. This can be simulated by jsbsim and Cannstatt bei propably implemented by me. Thanks to Ronald again!

HHS81:
To make sure: Oil cap: not on the Top of the engine cowling but on the side.

HHS81:
Ronald = Ron!!! F*** automatic correction on my Smartphone

gilbertohasnofb:
> The oil cap is on the right fuselage in front of the door. It is drawn on the texture and included by the normalmap I think.

Oh, I see it now! Perfect :+1: 

> I can add a hotspot but not today. 

No rush

> Please consider a use of different types of oil. This can be simulated by jsbsim and Cannstatt bei propably implemented by me. 

Hmm, I don't know how that would work then. In the c172p, I implemented it using a Nasal function. Basically if the engine is on, the oil level goes down (depending also on RPM). If oil level very low, engine coughs, if critical then it fails or can't be started. 

If you want something fancier than that than I might not be the right person to work on this issue. But if you don't mind me asking, do you really think we need different types of oil? I personally don't see the need.

HHS81:
> Hmm, I don't know how that would work then. In the c172p, I implemented it using a Nasal function. Basically if the engine is on, the oil level goes down (depending also on RPM). If oil level very low, engine coughs, if critical then it fails or can't be started.

Nothing wrong with (beside using nasal here - nasal really needed for that?)
[...]

PR #187, opened 2017-11-24
Oil management

hbeni:
Fixing #125 with extended oil simulation depending on time the oil was in service (so simulating need for oil changes)


HHS81:
@hbeni @gilbertohasnofb 
I didn't found time the last months, and the last weeks showed me that I can't work anymore late into night, since I have to get up early for job.

So it seems I missed a lot, and currently I'm looking through all files and check things. It will need some time today.

hbeni:
Updated the liverys (see also PR at the texture repo)

gilbertohasnofb:
> So it seems I missed a lot, and currently I'm looking through all files and check things. It will need some time today.

@HHS81 No rush. Just please don't miss my request to add @hbeni to the https://github.com/HHS81/c182-Textures repository, as he is now also working with textures.

HHS81:
@hbeni Great contribution!

HHS81:
@hbeni  
A nice method for oil viscosity controlled by temperature by Ron Jensen (jentron): [https://sourceforge.net/p/jsbsim/mailman/message/30264774/](https://sourceforge.net/p/jsbsim/mailman/message/30264774/)

HHS81:
@hbeni so we wouldn't need a logic like engine-killed. Just a simple function, where oil-level >certain level leeds to a higher frction and with that to a high `propulsion/engine/friction-hp` which stops the engine, more realism. That's btw. how I will implement prop strike...

`

hbeni:
Sounds good to enhance the sim with this sometime!
Can we get realistic numbers for jentrons "made up" numbers?

hbeni:
Oh, and the engine-killed property (and logic) would still be needed for two other reasons:
1. To support engine killing if water was in the fuel tanks (fuel-contamination)
2. To be able to expose some API to promptly kill the engine from outside (tutorial-code, telnet, ...)

That said, simulating friction would be cool nonetheless.

PR #189, opened 2017-11-26
Issue188

HHS81:
-Fix too low Idle RPM
-Fix too high Max Prop RPM
-Fix too high engine power


hbeni:
@HHS81 i will review this soon!

hbeni:
I readded the "oil-cold-effect" to the system; when you test, just disable the "complex engine" checkbox in Aircraft Options, this will disable it.

The numbers look better at a first glance, however i was not able to check this. I don't understand the lycoming engine chart (p. 3-40) fully yet.

hbeni:
OK, the inHG Numbers for the full-throttle with max-power-mixture seem good at all altitudes.
I cannot decide on the horsepower and the overall engine feeling as i do not have real world experience.

With the stated mixture and throttle, i think the oil gets somewhat cold - is this correct?
I cruised 10 minutes @6000ft at 120 TAS with the following power settings (should be somewhere at max economy) and oil temp was stable at about 100°F:
image

hbeni:
I'm still playing with the oil temp values.

hbeni:
i tested the cooling stuff in several altitudes and weather settings; turned out, i tuned the cowl flaps wrongly - settings seem to be OK!

hbeni:
@HHS81 looks good in my opinion!

HHS81:
@hbeni 

>  I don't understand the lycoming engine chart (p. 3-40) fully yet.
 ... 
I cannot decide on the horsepower and the overall engine feeling as i do not have real world experience.

You need the Cessna POH, not the engine manual! Pilots plan flights and check perfomance with the datas from the POH!

hbeni:
@HHS81 Ah thank you for that hint! I was not aware of those charts. I will recheck!

HHS81:
@hbeni 
Usually it is enough to place the aircraft at an airport at sealevel, change weather to 15°C, no wind and 1013Hpa. The atmosphere model of FGFS is quite good and delivers the correct temperatures for each altitude. 

Thanks for your test, I will check the numbers again.

hbeni:
(I just resolved the conflict to master: 2499346)

HHS81:
@hbeni @gilbertohasnofb 
Please check again, the numbers should now match the POH, I tested every altitude.
I also changed the thrust on idle, so taxi is is much more comfortable.
Also changed the prop animation, only below Idle rpm of 650 you will see the prop blades, above you will see the blurred disc. 

I think the aircraft behaves a lot better with this all!

hbeni:
Very good!
I will check the next days and report back!

Am 2018-01-05 0:11, schrieb HHS81:

hbeni:
@HHS81 against what value do you calibrate?
- jsbsim velocity vc-kts?
- instrumentation/airspeed indicated speed?
- instrumentation/airspeed true speed?

The Fuel consumption seems to be about OK from what i understand.

Against true-speed i still get off values:
- 6000ft@3°C / 19inHG / 2400RPM: 10GPH, 122 KIAS; 131 KTAS (KTAS should be 122)
- 6000ft@3°C /  23inHG / 2400RPM: 12.6, 127 KIAS; 139 KTAS (KTAS should be 136)
- 6000ft@3°C / 19inHG / 2000RPM: 8.7GPH, 112 KIAS; 112 KTAS (KTAS should be 109)
- 6000ft@3°C / 23inHG / 2000RPM: 10.65GPH, 134 KIAS; 146 KTAS (KTS should be 126)

How exactly do you test? and against which table?

hbeni:
With what i described above i get the following at 6000ft; i double checked weight and temperature.

Low rpm / Low inHG (KTAS should be 109, GPH ~8.7):
-> True speed too low / GPH too low
image

Low rpm / High inHG (KTAS should be 126, GPH ~10.8):
-> True speed too low / GPH little too low but OK
image

High rpm / Low inHG  (KTAS should be 120, GPH ~10.0):
-> True speed OK / GPH too low
image

High rpm / High inHG (KTAS should be 136, GPH ~12.7):
-> True speed OK / GPH too low
image

hbeni:
@HHS81 When i make mix a little richer, the values get closer, so i assume i was too near to the peak.
=> To conclude, i think it looks OK now from the small sample i tested with.

That was some hard work now completed by you! :thumbsup: :clap: 🏆

HHS81:
@hbeni 
Page 5-26ff of POH
Aircraft weight set 3100 and then sim/freeze/fuel set to true. Since the pages gives a weight of 3100lbs as reference.
The pages also gives true airspeed as reference, so take the one by FlightGear: instrumentation/airspeed true speed.
Detailed Weather, Advanced settings, Tile selection set to METAR. (gives accurate atmosphere!)

I lean as described in the POH, page 4-27: 

> _Lean the mixture slowly until the EGT peaks and begins to
drop._

But I'm not sure If did it right. Because it will also drop by inrichen the mixture

HHS81:
So I have to check again... *sigh*

hbeni:
@HHS81 i'm not sure either if i did the testing right. I have set weather to "fair" and then adjusted the METAR manually to produce the correct temperature.
I think also, i had the mixture leaned too much; as said above, if i increase mixture just a bit richer, (to align the GPHs used by the sim with those stated by the POH) numbers look also better here.

I think, that the numbers are "close enough" for the purpose of our simulation and _much_ better than in current master. If you ask me, we should incorporate the changes and accept it as "good enough" given the good numbers above, since testing this takes considerable time.

hbeni:
... seeing your commit i want to help you of course. Thank you for posting the test setup above, i will use this for reference now.
Please drop me a line once you are ready for the next test cycle.

HHS81:
Hmmm... While testing I noticed that while climbing fro 1000 to 2000 power increases to 250+horsepower.
There is something wrong with the table.

hbeni:
Where can you read out the horsepower?

HHS81:
fdm/jsbsim/propulsion/engine/power-hp

HHS81:
It will need longer- my notebook didn't like the "meltdown" security patch,  though having an Intel i5 Core.
I have to find out how to deal with it and how I can fix without too much work.

hbeni:
I wish you all the best and good progress!
Let me know if i can do anything.

HHS81:
Back online, everything running fine again.

Learn something new: Full Mixture and Full Throttle won't give us Full Power:
O&IO-540 Lycoming Engine manual p. 32

That explains a lot! 
New round of tuning! 😄

c182s.xml
```diff
@@ -1532,7 +1532,7 @@
     </external_reactions>
     
     <system file="c182s-stallhorn"/>
-    <system file="propulsion"/>
+  <system file="propulsion"/>
```
hbeni: @HHS81 btw how do you handle indentations in this project? I haven't figured out the policy from the code yet

hbeni:
..and you again deleted the dependency to the engines/engine/oil-powergain-factor.
This is used to simulate the engine don't have full power when oil is cold (currently, until the friction stuff gets coded)

HHS81:
I'm not ready yet, I haven't checked every flight level yet.
When changing flight levels engine power exceeded a lot.
I have disabled powergain as long I'm tuning all numbers.

HHS81:
Sorry,  but during the week I don't find time for FGFS. So as an example I'm just preparing an in-house training for nurses,  and other stuff not related to FGFS.

hbeni:
Okay thanks for the Information! 
No problem and no rush, just wanted to share my results with you in case they may help.

HHS81:
@hbeni 
Found some time today, could fix a lot of things and correct a long major issue. 😄 

1.) the correct lean technic according POH is to lean until the neddle reached maximum: 

> 1. Lean the mixture slowly until the EGT peaks and begins to drop.
> 2. Enrichen as needed to ensure operation at peak.

(p. 4-27) . Short said: keep it at the peak. Not below. I did it obviously wrong with my first versions here.

2.) I checked all flightlevels. The reason for the 6000ft mismatch was due a faulty leftover in the thrust table- horsepower was correct, but not thrust. At 14000ft speed is a bit too low (~2ktas), but that's o.k. I think.

3.) With commit `a968ce9` I had a massive increase in horsepower, exceeding 250+hp, while climbing. Found out that using fuel-flow instead of mp-hg doesn't let it increase and let all numbers be more exact. 

3.) With this method I was able to add _lean-for max.-power_. I found out, (and verified with X-Plane), that with full mixture and full throttle you won't have full power. The Engine manual says on p. 3-6, that with full mixture you only will have 95% available power. The POH states it as well: To get 100% available power you have to lean according the fuel flow placard on the left in on the panel (as an example p. 4-14; p. 4-24; ...). The reason why you still should set full mixture is [that it will protect the engines:](https://de.wikipedia.org/wiki/Leanen) 
That's something that seems wrong for many JSBSim aircraft with piston engine, especially the c172p and DR400.

4.) increased drag factor of the cowl flaps. Correct numbers now for max. climb with max. perfomance power according POH p.5-18.

5.) and I readded the oil-powergain-factor.

HHS81:
@hbeni 
I hope it should be all right now.

hbeni:
Some random numbers:

Sea level; looks good (6-7TAS too slow tough, but good enough!)
image

4000ft, high RPM (looks very close!!!)

4000ft, low RPM (all look well, but TAS too slow (-8), should have been ~125)

4000ft, low RPM with low inHG looks very close again (112TAS, expected 115).

6000ft, low RPM with low inHG is spot on.

6000ft, high RPM with low inHG is too fast (got 134TAS, expected 120; other values are good!)
image


6000ft high RPM with high inHG is good again!
image


8000ft, high RPM and low inHG is 11 too fast (TAS=132, expected 119)
image

8000ft, high RPM and high inHG is OK again
image

8000ft, low RPM and low inHG looks good (only 3TAS too slow, which is negligible)
image

8000ft, low RPM and high inHG is 6 TAS too slow (114 TAS, expected 122)
image


14000ft, high RPM, high inHG: i had to adjust to fuel flow heavily, with the fuel flow im way lean of peak egt. nonetheless we are too fast (129TAS, expected 118)
image
Leaned to peak-EGT i get 131 TAS at 10.42 GPH

14000ft, low RPM at 16inHG i get good numbers when leaned to GPH.
When leaning to peak EGT, i get 104TAS(spot on!) but 8.8GPH (expected 8.2)
image



What do you think of the numbers? I think we are pretty close; but with 6000 and above we are still a little off the numbers in some settings.
Good work so far! Please say, if the numbers are close enough for you.

HHS81:
@hbeni 
That was very helpful, thanks!

I can't get closer like this now, and I think it is good enough now.

hbeni:
Cool, i will recheck and report back.

hbeni:
More results (bold are numbers that are off in my opinion):

- **2000ft, 2400RPM, 25inhg:  131 TAS (expected 135)**
- 2000ft, 2400RPM, 21inhg:  121 (expected 120)
- 2000ft, 2000RPM, 25inhg:  125 TAS (expected 125)
- 2000ft, 2000RPM, 21inhg:  107 (expected 111 BUT OK since i wasnt able to closely match fuel flow)

- 6000ft, 2400RPM, 23inhg:  134 TAS (expected 136)
- 6000ft, 2400RPM, 19inhg:  122 (expected 120)
- **6000ft, 2000RPM, 23inhg:  130 TAS (expected 126)** (probably would settle below in some more minutes as trend was still _slowly_ declining)
- 6000ft, 2000RPM, 19inhg:  109 (expected 109)

- 12000ft, 2400RPM, 18inhg:  123 TAS (expected 128)
Speeds are off, but the leaning is very close now
- **12000ft, 2400RPM, 16inhg:  121 (expected 114)**
- **12000ft, 2000RPM, 18inhg:  128 TAS (expected 114)**

Speeds are good, also the leaning is very close now
- 14000ft, 2400RPM, 16inhg:  120 (expected 118)
- 14000ft, 2100RPM, 16inhg:  101 TAS (expected 105)



Very good work, i think we could leave it there.
Did you also check the intermediate altitudes, or should i conduct further testing for 4000, 8000 and 10000? Or is the interpolation be good enough, so we can trust it?

HHS81:
@hbeni 
I corrected the bold off numbers, and checked again all others levels. The interpolation is good enough, I didn't see wrong numbers at the intermediate levels.

Known: 14000ft, 2100rpm, 16inHg is still around 101kias, I can't fix that without influencing other altitudes

I don't know how exact the real values are, but I think we are overall not more away than 5%, which is quite good.

I also corrected the drag number of the cowl-flaps to match more p.5-19

Hopefully this time everything is right.....

hbeni:
Very nice!
Should i do a quick check again once i got some time?

HHS81:
Yep,  that would be great.

hbeni:
Ok looks still good now.
- 2000ft is okay
- 6000ft is okay

12000ft still off a bit:
- 12000ft, 2400RPM, 18inhg:  123 TAS (expected 128)
- 12000ft, 2400RPM, 16inhg:  100 (expected 114) --> **little too slow now; was too fast in last check**
- 12000ft, 2000RPM, 18inhg:  110 TAS (expected 114)  --> better, ok now

14000ft is also more off now.
- 14000ft, 2400RPM, 16inhg: 120 (expected 118)
- 14000ft, 2100RPM, 16inhg: 101 TAS (expected 105)


I would suggest to leave it there for the overall values.
Maybe you can revert to the old values for 12.000 and 14.000ft?

hbeni:
Hm, also the autopilot seems not to be as vertically stable as before, it always oscillitates, also on lower altitudes. Please check yourself, so we can rule out i did something wrong in the balancing.

HHS81:
@hbeni 

> Maybe you can revert to the old values for 12.000 and 14.000ft?

I can't, it would change the other altitudes back to the old off numbers. I can only try to fix 12000ft, 2400RPM, 16inhg.

14000ft numbers is btw exactly the same as in your test before, and above the altitude you have to use a oxygen mask..... 

> Hm, also the autopilot seems not to be as vertically stable as before, it always oscillitates, also on lower altitudes. Please check yourself, so we can rule out i did something wrong in the balancing.

It simply needs new tuned values, since the power settings now have been changed. And don't forget that the KAP140 for the C182 is only allowed to use at speeds above 80kias

HHS81:
@hbeni 
I got at 12000ft, 2400RPM, 16inhg: 118ktas

HHS81:
fgfs-screen-015

hbeni:
Have flown nearly 2hrs today and autopilot was ok.

I think this is ready :)
Well done, heiko, that was a lot if work!

HHS81:
@hbeni 
Thanks! Merged, and I will push now the existing master branch into FGAddon.

PR #191, opened 2017-11-29
Implemented basic aircraft startup options (fix #163).

### LINKED ISSUE #163: Start options
Currently the aircraft always start with the master and avionics ON and the throttle, prop and mixture levers at ready positions for a startup. I think we should we should add a radio dial with three options (only one of which could be selected) as for the startup:

* ready to start (which is what we currently have with the aircraft ready to start)
* cold and dark start (forces everything OFF)
* save state (save all switches, levers, etc. from the previous session)

hbeni:
They already work but especially engine start is not that nice implemented currently.


hbeni:
@HHS81 i think this is ready

hbeni:
@gilbertohasnofb look at the new auto mode :)

hbeni:
Also fixes #109

hbeni:
@HHS81 I will merge this if you don't throw in your veto ;)

HHS81:
@hbeni  It would be great if you could review my PR first. It is open since 5 days. 
And unfortunately I don't find much time under the week to review and work on due my daily job.

PR #197, opened 2017-12-04
Added winterization kit (decreases cooling airflow -> higher temps)

hbeni:
The winter kit reduces cooling airflow, helping keeping the engine warm at cold weather.

image

### DIFF
--- a/c182s-set.xml
+++ b/c182s-set.xml
@@ -528,6 +528,7 @@
             <path>/engines/engine/complex-engine-procedures</path>
             <path>/engines/engine/allow-fuel-contamination</path>
             <path>/engines/engine/allow-oil-management</path>
+            <path>/engines/engine/winter-kit-installed</path>
             <path>/consumables/fuel/tank[0]/level-gal_us</path>
             <path>/consumables/fuel/tank[1]/level-gal_us</path>
             <path>/consumables/fuel/tank[2]/level-gal_us</path>
@@ -780,6 +781,7 @@
    <allow-fuel-contamination type="bool">false</allow-fuel-contamination>
    <allow-oil-management type="bool">false</allow-oil-management>
    <oil-powergain-factor type="double">1</oil-powergain-factor>
+   <winter-kit-installed type="bool">false</winter-kit-installed>
   </engine>
  </engines>
  
--- a/c182s.xml
+++ b/c182s.xml
@@ -671,15 +671,16 @@
                   <max>1</max>
                 </domain>
                 <range>
-                    <min>0.48</min>
-                    <max>0.85</max>
+                    <!-- Defines cooling airflow with min=closed/max=open cowl flaps -->
+                    <min>0.35</min>
+                    <max>0.80</max>
                 </range>
                 <output>propulsion/engine[0]/cooling-factor-cowlflaps</output>
             </aerosurface_scale>
             <switch name="cooling-factor-winterkit">
                 <!-- winter kit reduces cooling airflow -->
                 <default value="1"/>
-                <test logic="AND" value="0.72">
+                <test logic="AND" value="0.50">
                     /engines/engine/winter-kit-installed  EQ  1
                 </test>
                 <output>propulsion/engine[0]/cooling-factor-winterkit</output>
--- a/c182s.xml
+++ b/c182s.xml
@@ -644,7 +644,7 @@
                  <output>/sim/model/c182s/cockpit/flaps-lever</output>
              </kinematic>
         </channel>
-<!--
+        <!--
             Cowl Flaps 
         -->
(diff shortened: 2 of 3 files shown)


hbeni:
Option can be enabled at aircraft options

hbeni:
@gilbertohasnofb Can you make this placard?

gilbertohasnofb:
@hbeni Sure, but what placard? Can you post a picture?

gilbertohasnofb:
@hbeni Ok, I understand what placard you are talking about. I will add it to the cockpit texture tomorrow.

gilbertohasnofb:
@hbeni It turns out the placard is already there:



The position is based on photographs of the cockpit, e.g.:



1b

HHS81:
@hbeni So it can be merged?

hbeni:
@gilbertohasnofb cool! I really overlooked that in the texture, sorry for bothering you without double checking!

@HHS81 I think it is mergeable; however what i'm still unsure about is the oil temperature. Maybe the current setting for the cooling-factor is too efficient, keeping the CHT too low at normal flight. I will play a little more with this okay? I will report back soon.

hbeni:
Ok, since 420fab3 i am much more satisfied with the temps.
Please note, that these temperatures are just guesses, i have no first hand experience.

@HHS81 I think you may merge this now, if you like

gilbertohasnofb:
@hbeni no worries!

hbeni:
@gilbertohasnofb Would it be a good idea to exchange the air-inlet texture when the kit is installed, so one can see visually in the preflight check that the kit is installed?
I don't know how to do this, however. Can you help?

gilbertohasnofb:
@hbeni I would be happy to help, but can you better explain to me what needs to be exchanged? I do not know how the winterization kit looks like and I don't know what needs to be changed texture-wise.

hbeni:
Hello Gilberto, 
unfortunately i was not able to find the exact c182 kit as images.
Basicly it consists of insulation (not visible) and metal plates with slits/holes mounted into the air intake behind he propeller, so the cooling airflow is physically reduced.

The 172 ones look like this: 172 baffles
The slits would be visible when one looks into the vents from the front.

hbeni:
Assuming from the POH-text (see first post) it may also be more like this: Two plates

hbeni:
Assuming from the POH-text (see first post) it may also be more like this: Two plates

gilbertohasnofb:
@hbeni I think I understand now what you mean by exchanging the texture: you mean that instead of creating a 3D model for the winter kit we should simply swap the air inlet texture, so one texture will have the kit and another won't. Is that correct? If so, I think this is not a very good route and I think we should go ahead with 3D modelling. Here are a couple of arguments:

1) it's much easier to simply add/remove a 3D object such as how we do with the wheel chokes
2) having two sets of texture for each livery would be an overkill and would really enlarge this repo
3) the air inlets in the c182s model are true holes (unlike the c172p which are made out of textures), so that the texture route wouldn't work IMO:



With that all said, if you or @HHS81 manage to come up with the 3D model, I would be happy to supply a texture. Would you prefer a dark plastic texture like the bottom one or a grey one like the first image?

Also, with a 3D object, we could make that by clicking on the nose cone the user can add/remove the winterization kit.

hbeni:
I agree a 3d model is better.
We need @HHS81 for that, it seems i cannot learn 3d modelling.

gilbertohasnofb:
I am also terrible at 3D modelling... So if @HHS81 is up to create this model, I would gladly make a texture for it!

HHS81:
I have to see if there is still some place left on the uvmap. I guess we want to have the placard on the Nosecap Cover readable.

hbeni:
Just out of curiosity, would it be possible to add a seperate texture image file? Otherwise all liverys must be redone...

gilbertohasnofb:
@HHS81 If I understood correctly, there is no need for a placard any longer (it's already there in the panel), there is just the winterization kit texture.

hbeni:
@gilbertohasnofb There is also a placard on the winter kit itself which @HHS81 is speaking about

HHS81:
@gilbertohasnofb The Nosecap Cover has also a placard

HHS81:
@hbeni 

> Just out of curiosity, would it be possible to add a seperate texture image file? Otherwise all liverys must be redone...

Would not be wise....

gilbertohasnofb:
> The Nosecap Cover has also a placard

I see.

HHS81:
@hbeni @gilbertohasnofb 
I will add the winterkit 3d-model (aka Nosecap Cover) tomorrow, I think I do know how to add the placard without changing all liveries.

gilbertohasnofb:
Great, I look forward to it!

PR #199, opened 2017-12-05
Electrical fix of Issue#43

HHS81:
this PR fixes the electrical system:

no cranking, no lights and no avionics below 22V
battery capacity now also depends on outside temperature


hbeni:
d08603f in other branch (issue200) adds influence of external heater to heat battery

hbeni:
@HHS81 This looks good to me.

PR #202, opened 2017-12-08
Added placard text to baggage door GUI

hbeni:
Added the POH placard text to the baggage GUI.
image

I think it would also be great to make the placard resolution in the texture good enough to be read in game from the walker.


gilbertohasnofb:
@hbeni this looks really good, great idea to add it to the dialog.

> I think it would also be great to make the placard resolution in the texture good enough to be read in game from the walker.

I think that might be a bit difficult as the resolution would need to be really high in order to make it readable.

PR #210, opened 2017-12-15
Fix custom registration livery (#209)

### LINKED ISSUE #209: Problems with the custom registration livery
@hbeni Just a couple of issues and ideas concerning this livery:

- perhaps you could redo the nose and tail as there are some strange places:

- maybe the blue colour could be a bit less saturated, but perhaps that's only my taste

- the callsign is actually "floating" and not touching the sides of the plane:

hbeni:
Fixes #209 
- [x] fix painting
- [x] fix location of 3D text, it should be closer to the planes surface


hbeni:
Ready to merge in my opinion - please have a test!

PR #201, opened 2017-12-06
Issue200 (Implement external heat)

hbeni:
- [x] fixed #200, now priming is depending on outside temperature. To enable priming in cold weather, external heater has been added.

- [x] Model and sound is still missing, can somebody add the 3D model and sound?
The property in question is /engines/engine/external-heat/enabled

See https://github.com/HHS81/c182s/issues/200#issuecomment-349787118 for a picture

PR #214, opened 2017-12-16
Fix Issue211

### LINKED ISSUE #211: Magnetos move when directing walker
The key "s" is bound to move the walker as well as putting the magnetos into start position.
The magnetos should not be moved, when one of the walker views is active.
The fix would be to enhance the "s" keybinding nasal script to check for the view.

hbeni:
This fixes #211 (magnetos move when walker walks backwards).
Note this patch is based on branch #201 (implement external heat), so should be added one we are ready there.


HHS81:
Please keep different issues seperated. This makes Reviews easier and prevents confusion.

hbeni:
thanks for the hint, the code relied on code in the other PR however.
How can i separate thie better?

PR #220, opened 2017-12-20
Dme ident volume fix

hbeni:
When the DME unit power-switch was switched to off, but GMA panel DME-Ident is on, the ident code was played.
I think the Ident should stop, when the DME is switched off, however.

This PR introduces a small nasal handler to the existing DME code that handles this.


hbeni:
sorry, wrong base

PR #221, opened 2017-12-20
Made DME ident stop when DME power is switched off

hbeni:
When the DME unit power-switch was switched to off, but GMA panel DME-Ident is on, the ident code was played.
I think the Ident should stop, when the DME is switched off, however.

This PR introduces a small nasal handler to the existing DME code that handles this.

### DIFF
--- a/Nasal/kn62a.nas
+++ b/Nasal/kn62a.nas
@@ -81,3 +81,13 @@ setlistener(nav1_state, func {
         }
     }
 });
+
+# Switch off ident volume when power button is off
+# (failing DME and loss-of-voltage is handled otherwise)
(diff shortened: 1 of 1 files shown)


HHS81:
DME ident? What's that? The morse Code you can actually here when Tunesien to NAV?

HHS81:
Here is the answer: https://www.avweb.com/news/avionics/183230-1.html

Do we simulate this?

hbeni:
Yes, we simulate this; it is built in as far as i have seen. The ident code already mutes when switched off at the GMA panel and also when power is lost on the DME.


> when Tunesien to

Just revealed that you are from germany :)

HHS81:
I don't like the Auto Correction by my Smartphone. 😀

PR #222, opened 2017-12-20
Remove tiedowns when in the air

hbeni:
Currently we can have strange effects when tiedowns are in place and the plane gets in the air.
This can happen for example, if the user presses STRG+U when tied down or (possibly, however already accounted for!) at in-air-starts.

When the Plane is more than 10ft in the air, the tiedowns are automaticly removed now.

(A short test just for fun showed that we need ~65 knots 90° to make the wind force dominate the tiedown force -> e.g. plane lifts off and tiedowns tear shortly after that)


hbeni:
I tested this now with several situations and it looks fine to me.

PR #225, opened 2017-12-23
Fix #223 (Fuel gauge always "on")

### LINKED ISSUE #223: Fuel gauge always "on"
I just noticed the fuel gauge always showing content. Is this intentional?
The POH tells "master switch on" then to check the fuel gauge - which leads me to the impression that it is electronically and should be implemented like in the c172p.

hbeni:
Fixed the fuel quantity an oil-tem/press gauges to not show indication without power.

@HHS81 @gilbertohasnofb 
Before merge i have questions:
- Should the VAC EGT and CHT gauges behave the same? (i think so, as they surely are electrically driven too?)
- What about ManPress? I suppose (like in "guess") this is mechanically driven?

### DIFF
--- a/Models/Instruments/ManPressFuelFlow.xml
+++ b/Models/Instruments/ManPressFuelFlow.xml
@@ -47,7 +47,7 @@
    <animation>
         <type>rotate</type>
         <object-name>FuelFlowNeedle</object-name>
-        <property>/systems/fuel/indicated-manfold-fuel-flow-gph</property>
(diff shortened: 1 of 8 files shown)


hbeni:
-For the VAC POH page 7-46 does not indicate electrical wiring.
- FuelFlow seems electrically driven (p. 7-21)
- CHT/EGT are also electrically driven (p. 7-20 / 7-21)

I will add that shortly.

hbeni:
Looks good now, please test :)

PR #228, opened 2017-12-25
Fog and frost (fix #192)

### LINKED ISSUE #192: Implement fog and frost

hbeni:
Enhances the simulation with fog and frost (fix #192 )

Fog and Frost has been ported from the c172p project.
I also added the defrost feature.
The idea is, that once the windshield is foggy, we can pull out "defrost" button and the frost/fog should get away more quickly. When we let the nob inside, it should get less at the same speed as in the c172.

Currently i had not much luck producing fog with proper fog settings, i cant reproduce the fog in real situations. For quick testing i added a (local on my PC) offset variable which allows me to tiner with the windshield dewpoint variable.
@gilbertohasnofb How can i produce realistic fog conditions for better testing?

Please also note, i cherry-picket the two commits from the breakers branch (break up panel object to get individual breaker switches) to avoid conflicts after merging that branch. ***This branch must be merged AFTER the breakrers brach***; and when doing so, we need to select this branchs interior-model.ac!


HHS81:
There is no frost visible, it looks like the interior glass is completely missing

HHS81:
O.k., the entry for the shader is missing

hbeni:
Today i conducted a flight from LFMK to LEGE and encountered Fog build up shortly after establishing cruise. The defroster did work like i would expect it, so i think this should be "good enough" for now.

Thank you for merging the other PRs, i will resolve the conflict here shortly.

HHS81:
@hbeni After some more flights, I get everything, frost and fog. Still I'm not sure if everything is realistic. 

I expect when temperature is below 0°C outside and inside, and people are inside to see some frost on the windows soon on ground as in the air. Here I can't see something like that.

hbeni:
Yes, i am not seeing this either.
I think this is because the heating of the people increases air temperature so humidity never reaches levels suitable to generate fog. I too think this is not realistically.
However this must be present at the c172s as well, as the code is nearly the same.

HHS81:
@hbeni I just let the aircraft rest at the ground, at OAT of -8°C, with open doors. Inside surface temperature decreased below 0°C.  Cabin-humidity were below 60% and only slowly increasing to 95%. 
Though inside surface temperature and high humidity I didn't see any frost.

hbeni:
But with open doors, should we really see frost at the windows? i would think that the humidity will mix with the outside air.
When closing the doors, i would expect frost showing shortly after however.

I just don't know with which variables to fiddle around currently. I will try to investigate further.

HHS81:
No, and that works pretty good. Frost/ Fog on the windows and opening the doors, and it is magically gone. I just let the doors open to get the temperature decrease fast. Then I closed the doors, but even with increasing humidity, nothing happened.

hbeni:
@onox i hope you are still around - can you please advise?

hbeni:
...Ok, i think i got lucky.

With the current change, frosting will start about 1 minute after closing the doors (and after temp/humidity accomodation of course);yielding first perceivable results at 1:30 and fully frosted windows after about 2:30 to 3 minutes.

hbeni:
... also some small thermometer inside the cockpit would be nice, so we get some feedback if its warm.
Or would a pilot self-comment would be enough, something like ("phuu, its wet inside" or "its hot here" or something?)

wlbragg:
Just a short comment here as I have no time to reply in depth at the moment. Part of the problem is that the weather parameters we use for the frost and fog are not correct at init time. They take as much as a couple minutes to "settle". So we start with 0 always which then has to correct to the actual temp and dew points over time. I am in the process of looking at this portion of the weather system and verifying that it is by design and not a bug.  So yes, this system is not set to be accurate at startup (such as sitting on the tarmac overnight and all frozen) which is what I originally intended. But instead it should always start out with no fog or frost and slowly adjust to the current conditions. There was a slight difference of expectations between what I started out with and what @onox programed as to the way the heater works. I think he has it where the heater can actually add to the fogging condition VS relieve it early on. I was trying to mimic the idea of a car window defroster that immediately would start removing a fogged windshield upon turning on the blower with heat. Whereas in @onox's implementation that can cause fog, see the warning about turning of the heater until the fog subsides. I never felt that was correct. But he did research this and technically the heat as in temp differential does or can cause the fogging. I think this is an area that can be improved. Also we do have the ability to simulate the defroster effect on the windshield. We haven't got to that part yet. I hope this helped clear a few things up.

wlbragg:
The c172p does give visual feedback as to the cabin temp as in something like "Temp is over 90deg" or Temp is falling below freezing" or at least it did at one point.

hbeni:
Thanks, @wlbragg for this hints.
I have seen (and copied) the feedback but think that the current scale is too coarse.
Maybe it would be nice to have a feedback based on temperature&humidity, something like a simplified [heat index](https://en.wikipedia.org/wiki/Heat_index)?
From that on, further simulation could be emerging, like getting a heat stroke or something :)

HHS81:
@hbeni 

> ... also some small thermometer inside the cockpit would be nice, so we get some feedback if its warm.
> Or would a pilot self-comment would be enough, something like ("phuu, its wet inside" or "its hot here" or something?)

I haven't seen a thermometer yet inside an aircraft. But those comments would be much better than "temp. reaching 30°C". We could code those comments to certain temperature ranges: 
"_My fingers are freezing_ " = below 0°C
"_A little bit fresh here_" = below 10°C
...
"_I feel good_" = between 18-25°C
"_It is getting hot in here_" = >25°C
"_Uh, are we taking a sauna in here?_" = >30°C

wlbragg:
I like it!

hbeni:
Ok, this should be easy to code up.

Howver, in the meantime, my change to adjust humidity reaching 100% instead of 95% by persons present in the cabin, seem to always make the windows foggy now, when doors and windows are closed - even in summer time.

hbeni:
I think i may have fixed this in b178a2a: now the cabin temperature also adjusts very quickly to OAT the first 20 seconds of the simulation.
Now i get:
- no fog in warm summer conditions
- fog in freezing weather (after ~4-5 minutes at -8°C with 4 persons)
The reason for fog (and not frost) is, that the cabin temperature raises above 0° due to persons aboard.

hbeni:
Ok, the text got overhauled now too.
When temp is in a critical range, (`if (temp <= 18 or temp > 30) `) it is reported every 30 seconds or so.
Otherwise it just prints when it changes the range.

Ranges are as following:
`if (temp < 0) {           setprop(txtp, "My fingers are freezing");

    } else if (temp <= 10) {  setprop(txtp, "A little bit fresh here");

    } else if (temp <= 18) {  setprop(txtp, "A little too cold here for my taste");

    } else if (temp <= 25) {  setprop(txtp, "I feel comfortably warm now");

    } else if (temp <= 30) {  setprop(txtp, "It is getting hot in here");

    } else {                  setprop(txtp, "Uh, are we taking a sauna in here?");

    }`

Any suggestions for the texts?

Otherwise i think the overall behavior is somewhat plausible now. I will make some in-flight tests and ask you to do the same.

hbeni:
I think, that a rough verbal estimate of the humidity would also be helpful, like 45%<>65% beeing good, other is too dry / too damp

hbeni:
What do you think?

image

HHS81:
You can't really feel humidity like you can feel temperature.

onox:
> I think he has it where the heater can actually add to the fogging condition VS relieve it early on. I was trying to mimic the idea of a car window defroster that immediately would start removing a fogged windshield upon turning on the blower with heat. Whereas in @onox's implementation that can cause fog, see the warning about turning of the heater until the fog subsides. I never felt that was correct. But he did research this and technically the heat as in temp differential does or can cause the fogging. I think this is an area that can be improved. Also we do have the ability to simulate the defroster effect on the windshield. We haven't got to that part yet. I hope this helped clear a few things up.

I didn't explicitly model any defroster functionality, just the general system of outside air influencing the cabin air influencing the windshield surface.

Furthermore there are 2 situations of the system:

1. How it works immediately after FG has started
2. How it works at any time t long past the initial settling time after FG has started.

Because the temperatures are set to 0.0 C, the settling time can be a couple of minutes. When @wlbragg is saying that he feels the system isn't correct, I hope he means situation 1, not situation 2 (taking into account no defrosters blasting air on the windshield have been modeled)

Some time ago I was reading the mailing list a bit and some JSBSim dev said you can initialize the <integrator> components by setting the input and output at the same time. In the C172p there are 4 of them: `heat/cabin-air-temp-degc`, `heat/surface-temp-degc`, `heat/cabin-relative-humidity`, `heat/moisture-norm`. If you set each input and output to the same value, you might fix situation 1.

@hbeni I wrote `heat/cabin-dewpoint-perception-norm` (0.0 .. 1.0) for that. It's a simple function.

onox:
Anyone on some FG discord server btw? I don't read my github e-mail very often.

hbeni:
Hi @onox, thank you for responding!
> Anyone on some FG discord server btw? I don't read my github e-mail very often.

What is that? (sorry for asking dumb questions)

> Because the temperatures are set to 0.0 C, the settling time can be a couple of minutes. 

I tried to fix that with the <c1> of the integrator being a higher number during the first seconds of the simulation. That seems to work reasonably well.

>  Also we do have the ability to simulate the defroster effect on the windshield. We haven't got to that part yet.

Currently i added a simple simulation for this, that seem to work "good enoug"; when the defroster is activated, it lowers the dewpoint of the cabin and raises the rate of fog/frost change, thus removing frost/fog relativeley fast. The effectiveness of the defroster is tied to the EGT tempeature, so this only works if the engine is on and providing heat.




> You can't really feel humidity like you can feel temperature.

@HHS81, You can, did you ever walk out after a rain shower in a hot country? Or the very humid days in summer time? You cant sense humidity like you do temperature, thats right; but i tried to respect this by the rather "broad" band.
Should i remove the reporting alltoghether? Or should i adjust the limits of reporting to more exteme levels (ok=10%-90% maybe; so >90 is "too damp"?)

D-ECHO:
> What is that? (sorry for asking dumb questions)
Discord is a communication tool, mostly designed for gamers, see https://discordapp.com, there are two (of which I know) FlightGear servers on discord, you can join them with an inviation link such as the one in this topic:https://forum.flightgear.org/viewtopic.php?f=10&t=31449

HHS81:
@hbeni  Happy new year!

> You can, did you ever walk out after a rain shower in a hot country? Or the very humid days in summer time? You cant sense humidity like you do temperature, thats right; ...

That's the point. You can feel temperature direct, humidity only indirect. And that is much more subjective than feeling temperature. I know people who don't have that much problems with very humid days, in the opposite like myself. And I never heard people say "The air feels comfortable" when they mean that the humidity is o.k. 
Most people will open the window in a car then when they see humidity: fog or frost on the windows. But not when they "feel" humidity - because they don't feel it (Especially not with thick clothings).

hbeni:
OK, i tought that it would be a good info for the user that the air gets dampy, so he can adjust the air- and temp lever accordingly.

I will inactivate the code, so it is in the sources but not active.

hbeni:
@HHS81 done :)
Happy new Year also for you!

wlbragg:
>**When @wlbragg is saying that he feels the system isn't correct, I hope he means situation 1, not situation 2 (taking into account no defrosters blasting air on the windshield have been modeled)**

@onox  Yes, situation 1.

As you said, we never modeled the blower on the windscreen so situation 2 I think is accurate overall except the blowing of warm/hot air across the windscreen to rapidly remove the fog from that portion of the windscreen.
At any rate it sounds like @hbeni worked around this by changing the speed of defogging with the blower on which isn't my favorite option as it really should only affect the area that the air is blowing on, not the entire cockpit. The entire cockpit glass should only change ,speed or time wise, according to the entire fog/frost system that @onox originally set up.

If you change the speed  or time it takes to remove the fog/frost from all the surfaces when the blower is on, then you effectively change the entire system. Instead, what you should really be after is when the blower is on you get a gradual reduction of fogging and ice on all surfaces. The surfaces close to the blower vents should more rapidly clear. That would be ideal and we do have the capability to do this.
Maybe I can make it a New Year resolution to finish this portion as I think the effect would greatly enhance the visual. 😄

hbeni:
Yes that would be a very good addition. I implemented it quick and a little dirty because i did not find any option to distinguish between the windshield and side windows.
However it is certainly doable, the twin otter does something similar with the wipers and rain.

HHS81:
Does the Cessna 172/ 182 have an defroster like the car, where air blows? Are these maybe the vents at the top of the winddhield?
If so, the interior windows must be uvwrapped and needs a texture with rgb-channel applied, as I understod the wiki.

hbeni:
i don't know. As i interpreted the manual the defroster blows air to the windshield and i assume it works like in a car - at least we are talking about the same concept, that is "hot air from engine"...

HHS81:
Hmm... the glareshield from those photos I can remind doesn't look like it comes from there. But I'm not sure. That's why I'm asking

wlbragg:
> Are these maybe the vents at the top of the winddhield?

I was alway under the impression that this is the case, but I don't know for sure.

>the interior windows must be uvwrapped and needs a texture with rgb-channel applied, as I understod the wiki.

You know, now that I think about, it I think I actually got that to work in the Cub (J3Cub/PA18 JSBSim versions).

HHS81:
@hbeni 
Can we merge this?

hbeni:
I'm done, if you like the results we can include it in master.
The optimizing regarding the blowe can be left open as exercise for the future in my opinion.

PR #234, opened 2018-01-01
State additions

hbeni:
- [x] Rewrite of state-init system:
  Now it is possible to init state from command line: add for example "`--prop:/sim/start-state=cruising`"
  This is very useful in combination with `--vc=120` and gui-launcher setting "on final approach..." to practise landings

- [x] Additions to state code
    - Refactored state-code a little
    - state "ColdAndDark" is now always before preflight (secured and with fuel contamination and without oil/fuel checked

- [x] Experiment with state-overlay, so giving "`--state=xxx`"-Option is supported too

### DIFF
--- a/Nasal/c182s-states.nas
+++ b/Nasal/c182s-states.nas
@@ -6,7 +6,16 @@
 # The idea is that the checklist-states define states of the plane after conducting the checklist.
 # Common functions are centralized in specific functions.
 # The autostart- and state-functions then provide "bundles" of those checklists and override items if needed.
+#
+# The initial state can be overwritten by command line by adding for example: "--prop:/sim/start-state=cruising"
+#
+# Author: B. Hallinger 2017/2018
 ##########################################
+
+
+# The supported states in state property
+var supported_states = ["auto", "saved", "cold-and-dark", "ready-for-takeoff", "cruising"];
+var state_property   = "/sim/start-state";
 
 
 ####################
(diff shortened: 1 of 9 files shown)


hbeni:
After 03556f3, one can use the `--state` switch to configure initial state. This overrides the already set GUI-Option in the aircraft-dialog.
The "`auto`" setting is already aware of the in-air state, so it is now very easy to start practising approaches; just use the "`--state=auto`" option and configure the "on final approach"-location setting.
Alternatively configure offset to an fix (like VOR).
Now the state initializer is also aware of the initial airspeed. If it is not set otherwise (`--vc=...` or lancher), and an in-air-start is configured, the speed is initialized to 100 knots.

**Example (Approach to EDDM):** `fgfs --aircraft=c182s --state=auto --airport=EDDM --offset-distance=6 --altitude=3500 --timeofday=noon`
**Example (Power-off Approach to EDDM):** `fgfs --aircraft=c182s --state=take-off --airport=EDDM --offset-distance=6 --altitude=3500 --timeofday=noon` (engine idles, `parking` state gives the experience of gliding with all ~~hope~~ power lost and blocked pitot)

hbeni:
@HHS81 ready for test!

What i wonder is, how should i document such additions? Enhance the readme? Add to the wiki? Both?

HHS81:
@hbeni V
Very nice feature! 
Yeah, documenting.... The wiki is one of the first place where people look for references, I think.

hbeni:
ok, i added it to the discussion page already. Once we prepare another fgaddon release we should migrate this over to the main page i think.

HHS81:
I guess when I fixed the engine power numbers, I will simply put the current state into FGAddon. Since FGFS releases every 3 months, that's too much work for an extra release process. Since I have commit rights to FGAddon, I can commit whenever we want to.

wlbragg:
It might be worth mentioning this to James on the developers mailing list so he is aware of another test of the "state" system. I like your approach, the way you are using "auto".

hbeni:
Done :)

PR #235, opened 2018-01-04
Issue #137 (instruments glare)

hbeni:
Fix for #137

Issue #137, opened 2016-07-29
Gauges illumination doesn't match panel


gilbertohasnofb:
The illumination of the gauges does not seem to match the illumination of the panel. See examples below:
- Ex. 1: panel is dark but gauges are very bright; they seem to be very detached from the panel
  ![c182-0](https://cloud.githubusercontent.com/assets/5700795/17233499/a0fcb3dc-5530-11e6-8bb8-340070aa442c.jpg)
- Ex. 2: now both panel and gauges are matching illumination and the whole thing looks good
  ![c182-1](https://cloud.githubusercontent.com/assets/5700795/17233500/a98f098c-5530-11e6-8379-5fb26060cea8.jpg)
- Ex. 3: once again the gauges seem brighter than the panel
  ![c182-3](https://cloud.githubusercontent.com/assets/5700795/17233505/af372306-5530-11e6-9afb-a82112eefb41.jpg)

HHS81:
noticed some days ago - no idea why...

gilbertohasnofb:
Maybe @wlbragg would have some ideas about it?

HHS81:
After a test I think it is the material setting itself. (too much specular)

gilbertohasnofb:
Is the shadow effect applied to the gauges as well?

HHS81:
no, as I understood Thorsten it sin't possible as they are not part of the ac-model.

gilbertohasnofb:
I think I once read that if the gauges have their absolute position defined in the .ac model as opposed to translations and rotations in the model xml file, that would be possible. Wasn't it @wlbragg who told me that?

gilbertohasnofb:
See: https://github.com/Juanvvc/c172p-detailed/issues/790

hbeni:
To reproduce:
`fgfs --aircraft=c182s  --airport=EDDM --runway=08R  --start-date-gmt=2017:06:01:14:00:00 --disable-fullscreen=true`

hbeni:
@gilbertohasnofb I tried to play with the material settings in branch [Issue137](https://github.com/HHS81/c182s/tree/Issue137) and i think it is better now.
What bugs me a little is, that now the gauges seem to be too dark when no light is coming from behind. But on the other hand i was not able to fix the problem with specular settings alone; only reducing <ambient> and <diffuse> too did the trick.
Most probably i'm still not getting how this all plays togehter...


Here you see the "fixed" settings, when sun is behind aircraft (where the problem occurred):
[...]

PR #237, opened 2018-01-04
Added material animation to light marker lights in GMA

hbeni:
Additional fix for PR #235 (Issue #137): Added material animation to light marker lights in GMA.

Marker lights where dark in night with glareshield lighting on.


hbeni:
Checked it works both in test and in field mode on EDDM

PR #238, opened 2018-01-05
GMA-340: enhance marker handling

hbeni:
GMA340: Enhanced marker handling
 - marker is now silent, when audio panel is inoperable or shut off
 - marker can be temporarily muted now like described in manual (press "mkr-mute" once when marker-beacon is sounding)
 - Fix for lighting of beacon-indicators when GMA is powered off


HHS81:
Thanks!

PR #241, opened 2018-01-08
Add tyre spray effect (fix #240)

### LINKED ISSUE #240: Implement Tyre Spray and smoke
http://wiki.flightgear.org/Howto:Implement_generic_tyre_smoke

hbeni:
- [x] spray effect
- [ ] tyre smoke effect

### DIFF
--- a/Models/c182s.xml
+++ b/Models/c182s.xml
@@ -26,6 +26,74 @@
    </load>
 </nasal>
 
+ <!-- Tyresmoke and spray -->
+ <model>
+     <!--<path>Effects/tyresmoke/tyre-smoke-nose.xml</path>-->
+  <path>Aircraft/Generic/Effects/tyre-smoke-nose.xml</path>
+  <name>tyre-smoke-n</name>
+  <offsets>
+    <x-m>-0.2392</x-m>
+    <y-m>0.0000</y-m>
+    <z-m>-1.1000</z-m>
+  </offsets>
+ </model>
+ 
+ <model>
+     <!--<path>Effects/tyresmoke/tyre-smoke-port.xml</path>-->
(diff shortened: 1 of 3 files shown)


HHS81:
There is no tyre smoke on a real Cessna landing. To less speed on touchdown

hbeni:
spray and smoke working (default implementation)

(it rains cats and dogs here, but rainfall not visible on image):
image

A little rough landig at good weather:
image

HHS81:
Sorry,  but that doesn't look realistic. Have you ever seen a little cessna producing so much smoke like a big airliner? There is no tyre smoke on real landing. Can't be, to less speed on touchdown. 

And even while raining it is too much. And framerates.... 😑

hbeni:
@HHS81 thanks for the feedback, i just tought i quickly implement this with the hints on the wiki.
i notice no performance hit, but i didn't check either.

Should i investigate how to tune the water-spray effect down and remove the smoke alltogether?
Or should we leave that feature alone?

hbeni:
I removed smoke and tuned down the spray; how is this now?
image

hbeni:
just measured it. Crusing along in rain with the patch is about 21-23FPS here (and my machine is rather old), doing the same at the same rainy weather on master is 22-23; so a little faster at the lower band.
What numbers do you get?

HHS81:
fgfs-screen-004

still a lot of work until it is ready

hbeni:
> still a lot of work until it is ready

@HHS81 Hi Heiko, what exactly is missing / should be adjusted? (i think it looks good already)
If i have time the next few weeks i maybe can finish this off (we should not wait for the next fgaddon release i think; but maybe the current state is already mergeable enough for you; then i will fix the conflicts shortly)

PR #247, opened 2018-02-03
Some fixes after2nd release

### LINKED ISSUE #246: Several minor things after FGAddon Release 2
Today i found some minor things:
- [x] GMA marker beacons: strange behavior when no power (press test button -> they get darker?)
- [x] GMA ident should be off when no power is supplied anymore
- [x] Fuel quantity gauge indication is off a little (especially the right needle)
- [x] (optional) Fuel System / Tank crossfeed: This could maybe optimized more, currently there could remain 1% tank difference after uneven flight

hbeni:
Fix #246


hbeni:
Tests look good (i especially fear to confuse the autopilot with the tank crossfeed, however this looked good so far)

HHS81:
What was not correct with fuel qualitity gauge?

hbeni:
The right needle showed an off value (for example 5gals the needle was below that). But just a minor thing, no biggie.

Old: 
image


New:
image

hbeni:
I want to recheck that autopilot still works ok, before we merge this.
I report back!

hbeni:
Tested, looks good (inside the autopilot spec) and can be merged! 🎉

Issue #252, opened 2018-02-08
Add spiraling propwash effect


HHS81:
Dany93 from the c172p-team created athe propwash effect, which is obviously not simulated by jsbsim per default: https://github.com/c172p-team/c172p-detailed/issues/947

I know that X-Plane simulates it (quite heavy), so I think we should add this as well.

hbeni:
fixed by e22876b3154d751af797258605e4f037df452f41

PR #253, opened 2018-02-08
Issue #252: add propwash effect by dany93

HHS81:
This PR adds the propwash effect by dany93.
This means at full power and low speed you will need a signifficient amount of right rudder to keep the aircraft straight.


HHS81:
has anyone tested it?

hbeni:
no...

hbeni:
ok, now; what a strong effect :)
It makes starting really a challenge, especially without rudder pedals.
If i am allowed to propose, i think it would be advisable to make this realism grade user toggable in the aircraft menu.

HHS81:
Hmm, I thought this as well.
But the c172p has this enabled per default, and X-Plane of course as well.
And since the C182S is something more complex (IFR, constant speed prop, faster) I think that would be o.k.

hbeni:
okay then!
However i also cant tell if the effect is realistic; but lets assume this for now.

HHS81:
waiting to be merged

PR #255, opened 2018-02-09
Issue#249: prime improvement in cold weather

hbeni:
I'm fine with this so far :)

How do you want to have the freezing implemented? Should we do it simple like in the fuel contamination, where we just calculate an arbitary chance per hour that the freezing occur once the temperature drops below 6°?


HHS81:
I'm not fine with it yet. FlightGear doesn't like `Number°C` in the outcommented section, but I couldn't push the change yet. That's why I didn't open a PR yet - PRs only then when it is tested and ready to merge. 

The question is: When would spark freezing really appear?
From the Lycoming document, and this one here: [http://www.oregon.gov/aviation/docs/tips_on_winter_flying.pdf](http://www.oregon.gov/aviation/docs/tips_on_winter_flying.pdf)  it only appears, when the engine temp is below 0°C, and the engine only fired a few revolutions and then quits:

> Another  cold  start  problem  that  plagues  an  un-preheated  
engine  is  icing  over  the  spark  plug  electrodes.    This happens  when  an  engine  only  fires  a  few  revolutions  and  then  quits.    There  has  been  sufficient  combustion  to cause some water in the cylinders but insufficient combustion to heat them up.  This little bit of water condenses on the spark plug electrodes, freezes to ice, and shorts them out.  The only remedy is heat.  When no large heat source  is  available,  the  plugs  are  removed  from  the  engine  and  heated  to  the  point  where  no  more  moisture  is present.

HHS81:
Now it is working and I'm fine with

HHS81:
waiting to be merged

PR #259, opened 2018-02-14
Issue248

wlbragg:
Fixes #248 

This is finished as far as I am concerned and ready to merge.

PR #261, opened 2018-02-16
Issue248 flashlight effect on annunciator panel and immat

hbeni:
Fix of immat and annunciator panel flashlight effects path (see https://github.com/HHS81/c182s/pull/259#issuecomment-366349262)


hbeni:
looks good on my box.
@wlbragg can you confirm? I would merge then.

wlbragg:
Yes, this is fixed, good catch, thank you!

PR #263, opened 2018-02-18
Radiance correction for relative instruments

wlbragg:
This should fix the break to the relative instruments caused by introducing the shadow cube.


hbeni:
I will review soon.

hbeni:
@wlbragg Looks not so good unfortunately...
Is there a possibility that the effects detect if its "dark enough" (or the flashlight turned on?) so the effect only takes place when its night? This would disable the flashlight at dusk i suppose? Maybe a possible workaround...
What clean solutions do we have here?

Here a comparison (sun is in the back, instruments light):
image

Compare to the expected (this one directly from merge-commit 6feb09a; see https://github.com/HHS81/c182s/pull/235 resp. #137 ): 
image

wlbragg:
I'm looking into this!

wlbragg:
I understand what it is your seeing, but it is the correct lighting based on what is happening to the environment. When the sun is behind the aircraft then it is illuminating the gauges. At high noon the light on the gauges matches your material settings. Your artificially darkening them with the material animation I guess to make it fit better with the shadow cube? The correct fix is to position the gauges to absolute positions and do away with the material animations and allow the ALS system to do its job. So I guess it is up to you to decide what you would rather have. If you want the exact material animation you are using then you will have to get rid of the flashlight and ra-radiance.
Sorry, didn't realized what you were using the material animation for. If you really want the material animation settings then I don't think you want the flashlight because if you remove the flashlight effect from the gauges it's not gong to work. However you could still use the ra-radiance on everything except the gauges and I don't think anyone is going to notice it's not applied to everything.

hbeni:
The Material animation i did was a workaround in the perception that other solutions are not yet ready.
If changing the instruments location to the correct position is enough, i think we should go down that way. We had this already mentioned here in #137 / pr #235, referencing absolute instruments position in c172 (https://github.com/Juanvvc/c172p-detailed/issues/790).

What needs to be done now?
- [ ] drop material settings in instruments
- [ ] correct instruments positions in individual .ac files
- what else?

wlbragg:
>what else?

either use cabin.eff then for the instrument's effect if your planning on introducing the light-maps to the gauges after you make them absolute
or
add shadow cube back to c182s_interior_radiance and continue to use that effect for the gauges
 
>And would absolute positions reenable the (then correct) shadows inside the gauges?

I would test that theory by repositioning one of the vor's and maybe the altimeter and hook up the shadow cube again and make sure it is what we want and acceptable.

We are planning to do the same thing in the c172p.

HHS81:
What else?
remake the animation for each Instrument. 

I thought about it and I'm against it! The instruments wouldn't be reusuable any more for other aircraft. 
The radiance effect itself isn't physically correct,  the effect itself very subtile and hardly noticeable, especially on the already lit (!) Instruments.  I would rather skip it for the instruments then having such a surgery.

wlbragg:
>I thought about it and I'm against it! 

Your call, you can't do the flashlight either then. It's material animation the way you have it and no flashlight or any effects on the gauges or adjust the material animation to something you can live with or don't use the material animation.
 
Your using the material animation for one reason only, that I can tell, and that is to darken the gauges in one specific lighting condition, (sun on the horizon or at the back of the aircraft). It is not really doing anything else as far as I can see. Is that because you think the gauges are to bright during that one lighting condition? Is that because your casting a shadow on the panel and not the gauges and it appears to be darker at that time so the gauges appear too bright? So you want the shadow, but you don't want to live within the framework to get it on everything (absolute positioning), your fudging it.

The argument is, you can't reuse the gauges if you go absolute? You could use that argument and settle for many other reusable objects in an aircraft besides just the gauges, but typically we don't. I personally would rather loose re-usability than loose effects. But that is just my opinion. I never cared for shared resources as you have to put up with these unnecessary limitations.
Also, by not using absolute and insisting on a material animation you can not use any other existing effects on the gauges, no light maps, no shadows, no flashlight, no ra-radiance and no other effects. Your willing to use these effects everywhere else except gauges because of re-usability.

Just to be clear...
You'd rather mix and match effects that weren't designed to go together.
You'd rather use one old effect (material) and loose out on all other "newer" more immersive effects.
You'd rather do it the historical way and use XML to position only the gauges in the cockpit and no other object to claim re-usibility and loose out on all the new effects that are available.

I guess we see thing differently. I am in the other camp when it comes to this stuff. If you want these newer effects then you leave the old ones behind.

Like I said, it is totally your guy's call and I don't wish to influence you if you've already thought about all this and your OK with it the way you have it.

Especially since I haven't actually seen anyone use the shadow yet on gauges (as far as I know). I'm going to try it on the c172p, and if it works as I hope it will, then that is the way I would do it for any new aircraft. Until we have something else, it is what it is.

hbeni:
> Your using the material animation for one reason only, that I can tell, and that is to darken the gauges in one specific lighting condition, (sun on the horizon or at the back of the aircraft).

Yes, that was my approach to fix the glare issue within my means.

I liked the flashlight and also the shadow effect (if it would work correctly).
I could live without ra effect, but the flashlight is important for realistic night operation combined with electrical failure.

That said, i dont want to make a decision for heikos cessna, but would appreciate all the new effects :)

wlbragg:
>i dont want to make a decision for heikos cessna

Same here, it's his decision as far as I am concerned.

You both have worked hard to make it look this good and should be really proud of what you have already.

If it really is only about sharing of resources and that is the deal breaker, all these instruments are local to this aircraft anyway, your not sharing any of it. Unless your planing to at a later date.
If it is about the amount of work to position all the instruments then I can understand the reluctance. Although it really shouldn't be that hard as you already have the exact offsets needed. I guess I'll find out when we do the c172p.

hbeni:
I would volunteer to do the repositioning.
First just for some of the instruments to testnif we are on the right track, then if positive, all the other.

Heiko currently plans to add a c182t variant where the instruments would be shared as far as i see. But probably that should not be a big deal, the instruments would already be in roughly the right spot and could be readjusted with new offsets, if i understood right?

HHS81:
@wlbragg 
I agree to the argument about the instruments appropriate to the c182s due the markings.
And the VORs and the AI are already taken from Instruments-3d and so are more or less reused.

But this reuse is one unique feature of FlightGear, and many (more experienced) developers still insists on them. I can remember that some people wasn't happy when I started to improve the c172p 2010 and put the instruments in the aircraft-folder instead of linking it to Aircraft/generic/instruments-3d.

But there is another thing.
When I started the Cessna 182 3-4years ago, there weren't that much effects. There were only the 4-channel lightmaps. Then there came the interior-shadow-effects and everythings else..... But they are also more or less work-arounds, though sometimes quite impressive so far.

Great, but: if you want them all, you have consider their use and plan your 3d-work, animation and uvmapping. It got quite complex to incooperate them into an aircraft, and I guess that's the real reason why we do see hardly an aircraft which makes use of all them. Especially if you still want to support Rembrandt, you can get in trouble. I can't use the interior shadow effects on the EC 135 P2, since the ALS lightmaps aren't compatible with the one used in the default/Rembrandt-Renderer. And though I think Rembrandt is dead, I'm not sure if I should dump the support yet there. There are still a lot of people using Rembrandt....

I'm looking jealously to X-Plane. I have the demo, and I'm impressed with their graphics:
Shadows including self-shadowing, cloud shadows; PBR-materials (real time reflections, ...), real lighting.
Framerates are quite o.k. and compared with the amount of complexity of a scene they are equal or even better than FlightGear. 
I'm jealously because you just create the 3d-model, define the lights and position (no need to tinker with Blender that much) and create the normalmap. That's all. No creation of cubemaps, shadowmaps, seperate objects, etc....

With respect to Thorsten for his work - I do like the outcome. But the effect and graphic framework is much too complex, hardly flexible and misses a lot of features which are standard today.

And I have gained the feeling, by supporting it, it won't change. That's why my resistance.

HHS81:
@hbeni 

> Heiko currently plans to add a c182t variant where the instruments would be shared as far as i see

The older Cessna 182T is more or less the same like the Cessna 192 S. Boring :smile: 
I do the Cessna 182T with the Garmin 1000: https://www.flyperformance.com/wp-content/uploads/2015/04/N427AM-42.jpg or https://www.flyperformance.com/wp-content/uploads/2015/04/N956WM05.jpg
The later even still uses the KAP140....

HHS81:
@wlbragg @hbeni 
I'm still thinking about, I have not yet decided. But the argument, that the instruments are local anyway is quite strong, I must admit.

HHS81:
@wlbragg 
as long the instruments are simple "one-needle" gauges, and the panel is not tilted in any way, it is easy. 
But the AI as an example and more complex instruments are more difficult, especially when the panel is tilted in any way. 

If I would convert such instruments, then I would use the old Blender 2.49b and the FlightGear-export tool for by Melchior Franz: [fgfs_animation.py](https://sourceforge.net/p/flightgear/flightgear/ci/next/tree/utils/Modeller/)

wlbragg:
>especially when the panel is tilted in any way.
If I would convert such instruments, then I would use the old Blender 2.49b and the FlightGear-export tool for by Melchior Franz

It can't ever be easy! :frowning:

wlbragg:
@HHS81 @hbeni 

Update on shadow work I am doing on the c172p. I figured out we can instead of moving object to an absolute position, we can simply assign a custom eff file to the relatively positioned objects that moves the center location of the shadow cube VS moving the object. It works fine and after consulting with Thorsten appears to be no issue with overhead or taxing the system.

However I am still trying to get an acceptable solution to "rotation". As it stands now, even if we use this method to implement a shadow, the needle rotations or any other rotation in the instrument doesn't play nice with the effect (shadows and light appear at odd times). So it doesn't fix the lighting issue with any object that would be excluded from the effect. It does allow us to easily add shadows to inanimate objects or objects that only have translation. So even with an instrument you could get faces, bezels, etc. shadowed.

I'm still trying to get a better solution so you might want to keep this PR for a bit longer.

HHS81:
@wlbragg  still trying to get my Laptop back working. It Sounds like a good solution,  especially it could solve many issues on the Dornier 328.  I wonder why your solution won't work on animated objects. It does work on doors etc. correct,  does it?

hbeni:
That sounds good, and i'm very curious for your results :)

wlbragg:
>I wonder why your solution won't work on animated objects.

It's rotations that are the issue. I think the shader would need some kind of algorithm to calculate the cube face rotation when the object that it is applied to rotates.

hbeni:
Hi @wlbragg how is it going?

wlbragg:
@HHS81 @hbeni 

Update, i went as far as i could with this in the c172p. Turned out that while I could move the shadow cube to where it needed to be per relatively positioned objects, it was a major pain and glitchy to get the shadow cast to the correct position to match the rest of the absolute positioned objects and required a stand alone effect file for every object (although the light maps did as well).
All in all it was far easier to make all the instruments absolutely positioned and re-factor all their artwork and rotations. So that is what we did. When it came to the application of the shadow effect it ended up only being applied to fixed objects and a few that only had translations. anything with rotations such as needles had to have the shadow effect removed as it just doesn't work without the shader being programmed with that in mind.
Over all though, I think we ended up with a net gain in look and feel of the interior. A lot of work, but in my opinion, it was worth it.
A bit of pre-planning in a new project and this doesn't have to be anywhere as painful as it was for us.
Anyway, you can check out the master branch of c172p and judge for yourselves the results.

Feel free to close this issue if you still are of the mindset not to proceed.

It sounds like there is some rendering changes on the not to distant horizon, maybe that will give us some more and seamless tools to use.

HHS81:
According the forum, the new renderer makes good progress and we will get the same features of Rembrandt with the one from ALS. 👍

hbeni:
What does that mean for us exactly?

HHS81:
@hbeni hopefully interior and exterior shadows without any complex tricks....

HHS81:
@hbeni  @wlbragg  https://forum.flightgear.org/viewtopic.php?f=19&t=35285#p341789

wlbragg:
Yeah, I saw that, nice!

gilbertohasnofb:
Those screenshots are stunning! Really hope this feature make it into the main code.

hbeni:
FG Newsletter says, in 2020.2 shadow maps will get available :)
http://wiki.flightgear.org/FlightGear_Newsletter_May_2020#In_the_hangar

HHS81:
> 
> 
> FG Newsletter says, in 2020.2 shadow maps will get available :)
> http://wiki.flightgear.org/FlightGear_Newsletter_May_2020#In_the_hangar

For those who uses the nightly builds already have:
shadows1
shadows2

That also means, that we can now already animate the interiorlights much better: so you can finally make the ceiling lights moveable.

hbeni:
Wow, looks really great!
Do we need to make adjustments or does this come out of the box?

HHS81:
@hbeni 
I did some adjustements by using the NoShadow-tag to the light, and the use of the Interior-tag. 
So it comes now even better out of the box than before 😄 .
For the changes to interior lights I need some more time

HHS81:
https://sourceforge.net/p/flightgear/mailman/message/37160506/
We need to enable this in the effect-setting

hbeni:
Wow that will be so looking good!

hbeni:
We could do the outer lights as well with the compositor :)
That will be nice, doing preflight in night time :)

HHS81:
> 
> 
> Wow that will be so looking good!
> Are your changes already reflected in this branch?

We have to set <shadow-mapping-enabled> in the inherited Effect parameters, that is something new. I think all other things are already in the master branch

HHS81:
> 
> 
> We could do the outer lights as well with the compositor :)
> That will be nice, doing preflight in night time :)

The lights aren`t working yet in the compositor

HHS81:
@hbeni I have removed the radiance effect from the instruments, since it doesn´t really work with compositor. And I have updated the lights- so I guess I can close this pull request here?

hbeni:
Could not reproduce after you merged your latest copositor changes! :+1:

PR #269, opened 2018-04-08
No shadow when aircraft is upside-down (fix #250)

### LINKED ISSUE #250: Disable ALS shadow on upside-down crash
The ALS shadow breaks when the aircraft is upside-down. Best to disable it when aircraft is in that condition.

wlbragg:
Fixes #250 

Implemented this on the c172p as well.

### DIFF
--- a/Models/Effects/shadow.xml
+++ b/Models/Effects/shadow.xml
@@ -32,6 +32,9 @@
           </not>
           <property>/sim/rendering/shaders/skydome</property>
           <property>/sim/rendering/shadow-volume</property>
+          <not>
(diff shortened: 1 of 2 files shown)


hbeni:
Just tested it, looks good to me!

More favourable would have been if the shadow would be correct but this seems to be impossible with the ALS solution?

wlbragg:
Actually I see no reason the shader couldn't account for pitch and roll and reverse the shadow to work upside-down. Might not even be all that difficult to do if you know some math and GLSL coding. Pain to gain it may not be worth it though. By the time your upside down you really don't have much flight time left to get much use out of a shadow. Unless your doing acrobatics.

PR #266, opened 2018-03-28
Extended failures (fix #242)

hbeni:
- [x] Implements custom failures. All major components and many minor ones can fail now.
There is a (i hope userfriendly) GUI to trigger the failures and also with two surprise-modes. The GUI also has quick access to the basic compat failure modes.

Random failure modes can be initialized by setting properties from the launcher, making the flight (and/or preflight) more funny.
```# Init failure modes in case requested by startup.
# This can be done by providing the following properties through the laucher or cli:
#  + random failures (fail at most n systems in max x time)
#     --prop:/sim/failure-manager/surprise-mode/ammount=<number>
#     --prop:/sim/failure-manager/surprise-mode/maxtime=<minutes> (optional, default 30)
#
#  + random surprise mode (fail one system every x minutes): 
#     --prop:/sim/failure-manager/surprise-mode/timer-active=1
#     --prop:/sim/failure-manager/surprise-mode/timer=<minutes>  (optional, default 30)
#
#  To also turn off screen messages, use:
#     --prop:/sim/failure-manager/quiet=1
#
```

fix #242 

- [x] add user switch that allows restoring failures from a previous session

Issue #264, opened 2018-02-26
Implement proper oil friction Simulation


hbeni:
Just a friendly reminder for me to implement oil friction the right way sometime in the future.
Mentioned in https://github.com/HHS81/c182s/pull/187#issuecomment-355768921

- [x] The oil simulation could easily be leveraged to support failure of the engine driven oil pump (see failures issue #242 / pr #266 ) -- done in PR #270

hbeni:
Hi, i tried to adapt the example code in commit 93ba35d in branch `properOilFriction-issue264`  but i don't think this is quite right...
The POH states "FULL THROTTLE" but if i apply it, the engine quickly falters because of too high CHT (caused by too less airflow). It works if i advance throttle slowly (ie. balancing throttle against CHT), however.

The Short Field Takeoff procedure does _not_ work anymore (full throttle with brakes on, then leaning to obtain max power fuel flow according to placard value), the new engine CHT limit kicks in, applying friction and thus killing RPM and engine power. The max power can only be obtained with the normal takeoff procedure  combined with slowly advancing throttle.

What is wrong here? Should the engine tolerate high CHT for longer? (like a minute oe so?)
Or should we raise the upper CHT limit a little for the immediate effects? (CHT >500 should not have immediate effects on the engine or its power, but increase long term war greatly; shouldn't it?)

hbeni:
I don't like the experiments results. Also, the implementation seems rather complex to me.
Therefore i have come up with a simpler approach which lets us simulate things a little more easy which in turn allows better tuning.
As soon as i find time, i will implement the idea and upload a new branch for testing.

wlbragg:
@hbeni you tagged me, looks like 17 days ago, for this issue, I don't know if you meant to or not. I guess I wasn't paying attention to it as I never noticed it. I am not at all familiar with any of this code or the issue in general. I just wanted to let you know in case you meant to tag someone else for this issue. Good luck, your new approach appears sound.

hbeni:
@wlbragg i tagged you under the impression that you maybe know "the real thing" and give some advice. Thanks for responding back :) and dont worry

hbeni:
The implementation in branch https://github.com/HHS81/c182s/tree/properOilFriction-issue264-2 looks much better. I think this is already quite good.
Low oil condition is already ported to the new friction system, also.

hbeni:
Ok, with commit 48ec859 in branch `properOilFriction-issue264-2`, it should be mostly good now.
[...]

PR #270, opened 2018-05-29
Proper oil friction issue264 2

hbeni:
This PR is addressing/implementing the friction part of #264.
It will add another complexity to the engine system.
When the user has "complex engine procedures" active, he must obey to CHT limits or engine gets killed. If additionally "oil management" is enabled, also oil temp and oil pressure is of concern and must remain inside limits, or engine dies too. regarding this, the oil pump may die now by setting `/engines/engine/oil-pump/serviceable` to false (this is already supported in the failures branch/PR #266).

- The plane now calculates prop `/fdm/jsbsim/systems/propulsion/friction-damaging-engine`; it gets "1" very quickly if the engine gets abused (this may in the future also be used to simulate longtime wear&tear of the engine together with overhauling it). Drivers are mostly too high CHT and too low oil pressure (by lack of oil or failing oil pump).
- `/fdm/jsbsim/systems/propulsion/friction-critical` is triggered if damaging friction remains too long, currently about 1 minute. Once this border is crossed, the engine goes into eating itself in another 18 seconds, issuing black smoke and coughing. Only using the 'repair' button will help here.
- The engine-killing was previously implemented hacky by just cutting the fuel supply in `systems/fuel.xml`. Now this logic is ported over to the friction system. This way, if a engine-killing situation occurs (manually by kill-engine, or otherwise) we also have a nice spinning-down engine with smoke effect.
- The fuel contamination code has been ported too; no engine smoke is issued in this case.

I will now file a PR as i think this should be ready so far.
Please have a look at it if time permits :)


hbeni:
One thing we should look at is the oil temp friction.
Currently also low temperatures introduce some small ammount of friction.
Shortly after starting the engine, one must play a little with the throttle to keep the engine running; idle RPM is only enough if the oil is about 75°F or higher.
I don't know if this is realistic at all...

HHS81:
> Shortly after starting the engine, one must play a little with the throttle to keep the engine running; idle RPM is only enough if the oil is about 75°F or higher.

I Don't think it is realistic, but appreciate this contribution anyway!

PR #277, opened 2018-07-10
Made state overlays comply to new launcher

hbeni:
I corrected the descriptions text to comply to the launcher items.
No big deal, just documentatory.

the saved state was "broken" when called from command line (it was accidentally called "Saved" instead of the lowercase variant)


HHS81:
Thanks a lot!

PR #278, opened 2018-07-11
Fix #275: add drag when doors/windows are open

### LINKED ISSUE #275: Open windows/doors should add drag
Currently there is no difference between closed/open doors/windows.
Open windows and especially doors should add a bit of drag because of the bigger aerodynamic surface.

hbeni:
Currently i have the following, made up numbers (depending on density and speed; here at ~130kt):
- open door =~ 30Knots
- open window =~10 Knots

I don't know if numbers are realistic but the effect is noticeable and easily to adjust if needed.

### DIFF
--- a/c182s.xml
+++ b/c182s.xml
@@ -1026,6 +1026,43 @@
                     <value>0.028</value>
                 </product>
             </function>
+            
+            <function name="aero/coefficient/CDdoorL">
+                <description>Drag_due_to_door_left</description>
+                <product>
+                    <property>aero/qbar-psf</property>
+                    <property>metrics/Sw-sqft</property>
+                    <property>/sim/model/door-positions/DoorL/position-norm</property>
+                    <value>0.0250</value>
+                </product>
+            </function>
+            <function name="aero/coefficient/CDdoorR">
+                <description>Drag_due_to_door_right</description>
+                <product>
+                    <property>aero/qbar-psf</property>
+                    <property>metrics/Sw-sqft</property>
+                    <property>/sim/model/door-positions/DoorR/position-norm</property>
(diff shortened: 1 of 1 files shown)


HHS81:
Thanks.
With open Window it seems realistic. About the doors I'm not sure. 
I once read that you can steer a c150 by opening the doors, but I would guess that above some certain speed the doors can't be open due the air pressure.

HHS81:
I keep this open in the case we decide to simulate the blow down of the doors at certain airspeeds

HHS81:
https://github.com/c172p-team/c172p/issues/544

gilbertohasnofb:
@HHS81 Thanks, I will take a look at this code and import it to the c172p. Really appreciate you letting us know!

gilbertohasnofb:
@HHS81 I see now that this implements only the drag, not the animation. I am a bit stuck with the animation bit and had so little time to invest on it :cry:

gilbertohasnofb:
@hbeni I have used your implementation in the c172p (still did not create a PR) but I did tweak the values quite a bit. Opening the window is quite common in GA aircraft as far as I know, and so 10 kts is way too much drag IMO. I adjusted the window drag multiplier to `0.0020` and the doors and baggage doors to `0.0150`. You do feel a big difference but not that much, and once the animation is limited to open the doors just a bit when the airspeed is high then that will make even more sense.

hbeni:
@HHS81 i think we should use those numbers too, are you okay when i do this directly in master?

For the air pushing the doors, could maybe simulate this quite easily when clamping the door opening property to a value calculated by a lookup table depending on airspeed...

PR #281, opened 2018-08-08
Correct door drag numbers and force-close doors in flight (remainder from pr #278)

hbeni:
Open from PR #278 was the following:
- [x] use better drag numbers from c172p
- [x] slam shut doors above certain speed due to windforce
- [x] prevent door opening above certain airspeed
- [x] Fix drag propety for doors
- [x] produce sound when doors are closing fast
- [x] save door positions trough sessions. Make the doors close automatically in startup-modes (cruise, ready-for-takeoff, more?).

Should we port over the other two functions as well, or reimplement them in a simpler nasal approach, or skip alltogether?

### DIFF
--- a/Models/c182s.xml
+++ b/Models/c182s.xml
@@ -707,7 +707,7 @@
    <object-name>WindowFrame.L</object-name>
 <object-name>Window.L</object-name>
 <object-name>Window.L.001</object-name>
-   <property>sim/model/door-positions/DoorL/position-norm</property>
+   <property>sim/model/door-positions/DoorL/position-norm-effective</property>
    <interpolation>
      <entry>
        <ind>0.0</ind>
@@ -755,7 +755,7 @@
    <object-name>WindowFrame.R</object-name>
 <object-name>Window.R</object-name>
 <object-name>Window.R.001</object-name>
-   <property>sim/model/door-positions/DoorR/position-norm</property>
+   <property>sim/model/door-positions/DoorR/position-norm-effective</property>
    <interpolation>
      <entry>
        <ind>0.0</ind>
@@ -1096,7 +1096,7 @@
  <animation>
    <type>rotate</type>
    <object-name>BaggageDoor</object-name>
-   <property>sim/model/door-positions/BaggageDoor/position-norm</property>
+   <property>sim/model/door-positions/BaggageDoor/position-norm-effective</property>
    <interpolation>
      <entry>
        <ind>0.0</ind>
--- a/Models/interior-model.xml
+++ b/Models/interior-model.xml
@@ -155,7 +155,7 @@
 <object-name>doorhandleint_left</object-name>
 <object-name>FrameLock1.L</object-name>
 <object-name>FrameLock2.L</object-name>
-   <property>sim/model/door-positions/DoorL/position-norm</property>
+   <property>sim/model/door-positions/DoorL/position-norm-effective</property>
    <interpolation>
      <entry>
        <ind>0.0</ind>
@@ -190,7 +190,7 @@
 <object-name>doorhandleint_right</object-name>
 <object-name>FrameLock1.R</object-name>
 <object-name>FrameLock2.R</object-name>
-   <property>sim/model/door-positions/DoorR/position-norm</property>
+   <property>sim/model/door-positions/DoorR/position-norm-effective</property>
    <interpolation>
      <entry>
        <ind>0.0</ind>
@@ -264,23 +264,51 @@
     <animation>
         <type>rotate</type>
         <object-name>doorhandleint_right</object-name>
-   <property>sim/model/door-positions/DoorR/position-norm</property>
+        <property>sim/model/door-positions/DoorR/position-norm-effective</property>
+        <condition>
+            <equals>
+                <property>/sim/model/door-positions/DoorR/opened</property>
+                <value>0.0</value>
+            </equals>
+        </condition>
+        <condition>
+            <equals>
+                <property>/sim/model/door-positions/leftDoor/opened</property>
+                <value>0.0</value>
+            </equals>
+        </condition>
         <interpolation>
             <entry>
                 <ind>0.0</ind>
                 <dep>0.0</dep>
             </entry>
             <entry>
+                <ind>0.18</ind>
+                <dep>70.0</dep>
+            </entry>
+            <entry>
+                <ind>0.26</ind>
+                <dep>70.0</dep>
+            </entry>
+            <entry>
+                <ind>0.35</ind>
+                <dep>110.0</dep>
+            </entry>
+            <entry>
+                <ind>0.43</ind>
+                <dep>110.0</dep>
+            </entry>
+            <entry>
                 <ind>0.5</ind>
-                <dep>50.0</dep>
+                <dep>0.0</dep>
             </entry>
             <entry>
                 <ind>1</ind>
-                <dep>50.0</dep>
+                <dep>0.0</dep>
             </entry>
         </interpolation>
 		<center>
-			<x-m>0.969679</x-m>
+			<x-m>0.903840</x-m>
 			<y-m>0.569469</y-m>
 			<z-m>0.143912</z-m>
 		</center>
@@ -400,23 +428,45 @@
     <animation>
         <type>rotate</type>
         <object-name>doorhandleint_left</object-name>
-   <property>sim/model/door-positions/DoorL/position-norm</property>
+        <property>sim/model/door-positions/DoorL/position-norm-effective</property>
+        <condition>
+            <equals>
+                <property>/sim/model/door-positions/DoorL/opened</property>
+                <value>0.0</value>
+            </equals>
+        </condition>
         <interpolation>
             <entry>
                 <ind>0.0</ind>
                 <dep>0.0</dep>
             </entry>
             <entry>
+                <ind>0.18</ind>
+                <dep>70.0</dep>
+            </entry>
+            <entry>
+                <ind>0.26</ind>
+                <dep>70.0</dep>
+            </entry>
+            <entry>
+                <ind>0.35</ind>
+                <dep>110.0</dep>
+            </entry>
+            <entry>
+                <ind>0.43</ind>
+                <dep>110.0</dep>
+            </entry>
+            <entry>
                 <ind>0.5</ind>
-                <dep>50.0</dep>
+                <dep>0.0</dep>
             </entry>
             <entry>
                 <ind>1</ind>
-                <dep>50.0</dep>
+                <dep>0.0</dep>
             </entry>
         </interpolation>
 		<center>
-			<x-m>0.969679</x-m>
+			<x-m>0.903840</x-m>
 			<y-m>-0.569469</y-m>
 			<z-m>0.143912</z-m>
 		</center>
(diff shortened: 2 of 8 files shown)


hbeni:
@gilbertohasnofb I will - however i don't think it is "better".
What i would do is just a cheap hacky solution in nasal:
c182.nas would check periodically with a timer (every 2 or 3 secs or so) the airspeed and if the doors are open, something along (pseudocode):
```
if (airspeed > threshold and doorOpen) {
    interpolate(doorProperty, 0, 0.5);
    playSound("slamShutDoor");
}
```
This way the pilot can "try" to open the door but it will close very quick after that; that kind of simulates the pilot force-pushing the door but cannot hold it open due to windforce. This is, admittedly, not so nice as the current c172p simulation but "well enough" in my opinion for this corner case.
Another option would be to "lock" the doors shut with setting the property readonly if its closed an airspeed too high which would prevent the opening alltogether above certain speeds.

Also contrary to the discussion in the c172p issue, i don't think this is computational expensive, since those checks are only executed every other second (wich should be reasonable "close" for this minor feature!), which is every 240th FG engine tick. the XML-solution will run in c-space every tick as far as i understand the documentation.

gilbertohasnofb:
Cool, once you are done I will definitely take a look at it and compare to what I came up. Really curious on how this will look like!

hbeni:
I'm done. I think the effect is nice enough and forces the doors closed in flight.
When one opens the doors in flight, there may be situations where just the handle is pushed "closed" because the next loop iteration is too close. This is somewhat unrealistic, however, flightgear isn't a door simulator after all. If need really be, we could add a threshold to the current door position, so the handle will always be allowed to extend.

@gilbertohasnofb please see commit feaea5a for the nasal solution (containing debug messages for easier porting)

hbeni:
@HHS81 I just have seen at testing, that the door positions are not yet saved to the next session.
The reason is the initialization of the door objects. This would be easy to solve, we just need to make the init recognize the saved values and adjust the states code so to close the doors at initialization.

Should i? or is there a reason not to save doors position?

hbeni:
If performance is really so crucial that we need to save a few milliseconds every 2 seconds, we may enable/disable the timer based on a listener to the -nirm door position properties, so it runs only if the doors are not closed.

gilbertohasnofb:
@hbeni thanks for the heads up. I think this is a simple and elegant solution but I actually prefer what we currently have at the c172p, not due to implementation but just how it looks. I like that the doors are not fully closed in our aircraft and how they move in a non-symmetric way. But yeah, this is totally suitable as well.

hbeni:
@gilbertohasnofb can you port this over please? Just remove my code in this branch then :)

gilbertohasnofb:
@hbeni You mean porting my solution from the c172p to the c182s?

hbeni:
@gilbertohasnofb yes :)

gilbertohasnofb:
Sure, I will try to get this done today!

gilbertohasnofb:
@hbeni this might be a bit more work then I remembered. I am really busy using all my free time trying to get the c172p as ready for the 2018.3 release as we can, so would you mind if I would take a look at this after the release is done?

gilbertohasnofb:
@hbeni I decided to do this now as my TODO list is gigantic at the moment. It worked quite well, the only thing is that the c182s model needs a small adjustments for the correct animation of the interior door handles. You see, the pivot point of the door handle is not at the extreme back but rather in the middle. The place where the handle is needs to be 'carved', so that it can accommodate this rotation. See:

image

And compare to this video (from 0:20):

https://www.youtube.com/watch?v=kfGN_QEg1UU

But that's on the modelling side, which I can't do. So either the model needs this small modification or the x axis of the pivot point (just below the two new interpolation tables, one for each interior handle) needs its value reverted.

wlbragg:
I know this is way to late, but. I just pulled this and tried the last version
76d55919a46f468f0aface898164e8f14a75e5a5

That is really close to what I expected for this action. 

The exceptions being...
1) The door closing animation is too fast 
2) They close all the way
3) Airspeed may be a little too high before they start to close.
4) I would have added an animation to open them if there were enough deceleration g's VS wind resistance. Because I would not have them close completely like the c172p version does.

I think all the above wouldn't be that difficult to add to the 76d55919a46f468f0aface898164e8f14a75e5a5 version

Just my opinion and what I originally envisioned for this animation.

The final solution looks OK, I just cant get past the doors opening and closing being tied directly speed. It just isn't realistic. :heart:

gilbertohasnofb:
> The door closing animation is too fast

The only way of improving that would be to set a higher speed for the limit (when the door is nearly closed). It is a thin line we are walking since that would also mean that a door would remain a bit opened at much higher speeds. Right now I think they are fully closed at around 50-55 kts. The speed of closing also depends on the acceleration, since the rate of change of door position is proportional rate of change of speed. This is the best compromise I found but I would be happy to change it if anyone has a specific suggestion.

> They close all the way

They don't, they remain slightly opened. Zoom in and you can see that.

> Airspeed may be a little too high before they start to close.

Hmm, they start closing at anything > 0 kts so I don't understand that.

> I would have added an animation to open them if there were enough deceleration g's VS wind resistance. Because I would not have them close completely like the c172p version does.

I don't understand that. Why the c182s would behave different than the c172p, or am I misreading you and you suggest this change to both planes? If so, the trick bit is that the doors are effectively always opened in the current implementation, so there is no such thing as triggering an open command when deceleration is higher than a threshold. What you ask would require a completely different approach and as I mentioned before I just don't know how to conceive that with filters.

> The final solution looks OK, I just cant get past the doors opening and closing being tied directly speed. It just isn't realistic. :heart:

I sympathize but frankly I can't do better. Anything else would require a fancy system and a) I wouldn't even know where to start with and b) just don't want to spend more time with a minimal effect. But again, if anyone can come up with something better I would not mind completely removing my code from either/both planes at all! :smile_cat:

hbeni:
I really think we shouldn't overengineer this :)

wlbragg:
For clarification only, I think the filter solution ended up being the wrong approach for this effect, because of the limitations it imposes by tying directly to the airspeed and only the airspeed and it being a one-to-one cause and effect solution. That is just not how it works in RL.
 @hbeni former nasal solution (with a few more modifications) would more accurately mimic the real life effect wind and acceleration/deceleration would have on an open door.

I get it, using the filter for this was relatively painless and efficient and for all intense and purposes will suffice.

The only reason I brought this up again here was because I tried @hbeni early on approach and it was really close to what I had envisioned in my mind as to how this should work (with a few more modifications). But it is a nasal solution and I know we are trying to reduce nasal overhead. I just wonder if sometimes we don't settle for less effective solutions because of our hesitancy to use a bit of nasal.

>I really think we shouldn't overengineer this :)

I keep hearing this as well and I don't think that getting an effect accurate to the task it mimics is "over engineering". It certainly could be, but it doesn't have to be.

Anyway, I hope that clears up my comments, carry on. It's just one opinion and I felt it needed to be put on the table.

hbeni:
Yes, extending the nasal solution would not be much hassle. The speed etc would only be a single parameter. Taking wind into account would also be simple. Making it efficient so nasal only runs if it has to (when doors are open) would be simple too. Opening unlatched doors when braking should also be easy.
Probably only nasal can deliver all desired effects without creating a huge and very complex (to maintain) xml system.

I think @HHS81 should decide what sould be the final aproach as we three can live with both.

gilbertohasnofb:
@wlbragg The filter solution does have some nice things that I doubt you can create with a simple Nasal function without too much pain. For instance, the doors do not close completely, you can see that they are still a little bit opened, which is nice and fairly realistic while the Nasal solution was simply if airspeed > threshold then the doors were closed. The doors are also not closing synchronizedly, which is also very nice. Also, I find it really weird that the doors do not move at all until a certain airspeed and then simply close in the same way as when you close by clicking on it. But I do understand what is bugging you and I am happy to change it for whatever you feel that looks better.

wlbragg:
>But I do understand what is bugging you and I am happy to change it for whatever you feel that looks better.

@gilbertohasnofb this works well enough. I can't say I ever leave the doors open when I take off so I think the fact that they will react at all if a user does so is a bonus and will certainly surprise them. I doubt anyone but us are looking that closely at a particular animation and analyzing it as we do.

HHS81:
@hbeni @gilbertohasnofb @wlbragg 
Well, I like how it works in the moment, it feels realistic to me.

If we really want to have it that 100% realistic, we should assume the doors are aerodynamic surfaces and model it in JSBSim. But that would be something for dany93....

But I agree to hbeni that everything you adress could be solved in an easy way via nasal. 
On the other side this could be also done via filters. :thinking:  

>  Also, I find it really weird that the doors do not move at all until a certain airspeed and then simply close in the same way as when you close by clicking on it. 

It is not weird, actually it feels realistic to me. The doors has some weight and the door hinge has some friction - so it needs some air pressure to get the door closed. The more air pressure, the faster it closes. 
That should be also be able to simulate easily via filter. 

So, I like the current implementation via filters, and I think the "simply closing of he doors" can be solved with filters. Let me know when I should merge it.

hbeni:
Im fine with that. :)

hbeni:
Just did a test, it seems the baggage door stays open...

hbeni:
Ok, i investigated; the door *was* open but only after landing, like expected.

@gilbertohasnofb @wlbragg i noticed two other things:
1. The drag numbers look like "fully open", when the doors are just a slit open. It seems i (now) use the wrong property; which one indicates the current open angle, so i can tie the coefficients correctly?
image
2. When the doors close, they don't produce a sound (as they never reach the "closed" state). How (if?) do you produce the "slam shut" sound in the c172?

wlbragg:
@hbeni I think I can answer the door sound question if I am understanding your question correctly.
This is the left door closing sound trigger in the ci72p c172p-sound.xml file...
~~~
        <left-door-closing>
            <name>left-door-closing</name>
            <mode>once</mode>
            <path>Sounds/door-closing.wav</path>
            <condition>
                <less-than>
                    <property>/sim/model/door-positions/leftDoor/position-norm</property>
                    <value>0.5</value>
                </less-than>
            </condition>
            <position>
                <x>0.75</x>
                <y>0.6</y>
                <z>-0.35</z>
            </position>
            <reference-dist>0.2</reference-dist>
            <max-dist>50.0</max-dist>
        </left-door-closing>
~~~

hbeni:
@wlbragg Thank you! Am i right that this configuration does not issue a door sound when the door is forced closed by wind, since then the norm property is somewhat bigger than the defined treshold?

wlbragg:
Yeah, I think so if position-norm is >= 5 then no sound.

hbeni:
Is this intentional or a bug?

wlbragg:
That would be correct. We don't want door closing sound if the door isn't closing all the way, right?
Do you think it should make a closing sound, as in it is close to closing and maybe touching the frame so it should make a sound?

hbeni:
I wouldnassume

hbeni:
I think its better now :)

HHS81:
For me it is o.k.! I would merge it, any objections?

wlbragg:
Sorry I didn't see this until now.

>I would assume such an sound when the door is forcibly closed by air drag

My understanding is the door wasn't closing all the way and that is why there is no sound. If it is actually closing where and it makes contact with the frame then, yes, I think there should be sound.

hbeni:
@wlbragg I assume the door having enough force to produce a sound when it closes by wind as there is no "spring" forcing it open and working against that wind force. Having the doors closed without sound also feels a little awkward. i think audible response is also important for the pilot to advice of the issue.

hbeni:
Before merging this, i want to add the door position persistence between sessions, should be no big deal.

hbeni:
> Before merging this, i want to add the door position persistence between sessions, should be no big deal.

Done, and cleanup for merge into master.
It should be all set now.

HHS81:
Thanks a lot!

Issue #280, opened 2018-07-30
Ground equipement dissapears when aircraft is moving


HHS81:
Selected Ground equipement dissapears when aircraft is moving. Expected and feature was, that ground equipement will stay at its place even with aircraft moving, flying etc.....

Investigating.....

HHS81:
also the ground chocks doesn't dissapear. I guess something to do with the state-support. I hope that's inside our aircraft-model and not another FGFS-core problem.

HHS81:
I guess I know when the issue had been introduced....It seems not to be the state-support

HHS81:
O.k. it was the preheater.
The code should disable the preheater when speed is higher than 0.1knots. But with that it introduced the issue here. Since the code isn't needed anyway, since we already had a similar function in the gui and elsewhere I commented it out.

PR #285, opened 2018-08-09
Fix Issue #280

HHS81:
Fixed Isse #280 bis commenting out not needed code which introduced this bug


hbeni:
Confirm fixed and merged

Issue #296, opened 2018-09-08
Minor corrections


hbeni:
Today i had the opportunity to sit in a c182t and observed some things applying to the 182s too:

- [x] The davtron was "off" when battery switch was off, however the ET timer was persistent. Until now i thought from the manual that the little internal battery fused the whole device while in reality it seems that it only drives the quarz clock (like a CMOS battery in a computer). (done in branch for #296 because of dependencies)
- [x] oil pressure in idle is already inside the green arc. Oil pressure is not directly RPM tied and needs some brief initial time after start to show values (as mentioned in the manual, this is depending on the oil temp). We should simulate both effects.
- [x] Starting then engine (in warm weather with warm engine!) resulted in a small puff of dark smoke from both cowlings.
- [x] correct tooltip update at pitot and cabin heat/air levers

hbeni:
The oil-takes-some-time-to-show-values could be simulated quite easily with delaying the already existing oil pump -norm property; for example multiply it with a factor-mapping table depending on oil temp. This should simulate cold oil quite reasonably. The manual says that the pump needs some time to get pressure because of viscosity of cold oil.

hbeni:
Smoke is implemented; however currently still lacking texture. I don't know why, but asked on the forums: https://forum.flightgear.org/viewtopic.php?f=47&t=34771

wlbragg:
@hbeni I commented out all but the exhaust-smoke-left effect and called it exclusively in the Models/c182s.xml file

~~~
<model>
  <path>Aircraft/c182s/Models/Effects/c182_enginesmoke.xml</path>
  <name>exhaust-smoke-left</name>
  <offsets>
    <x-m>-0.60</x-m>
    <y-m> 0.00</y-m>
    <z-m>-0.10</z-m>
  </offsets>
</model>
~~~

You get your individual round puffs then. I think it is a combination of calling it all at once.

hbeni:
@wlbragg Fixed and implemented, it seems only one effect is allowed per definition file.
![image](https://user-images.githubusercontent.com/13608602/45304717-f6d81900-b518-11e8-9caf-194b94089ce7.png)

wlbragg:
[...]

PR #297, opened 2018-09-11
Issue296 minor corrections

hbeni:
see #296:
- [x] davtron is off now without power
- [x] oil pressure in idle fixed
- [x] cowl smoke briefly after start (mixture too rich because of start)
- [x] tooltips of pitot and air/heat levers each now update on both properties changing
- [x] even better tooltip showing current warmth status


hbeni:
Alan informed me on a new property: https://forum.flightgear.org/viewtopic.php?p=336544#p336544

I want to use that for calculating smoke, as this will also work in flight with generating exhaust smoke in vety rich conditions

hbeni:
Ok, it unfortunately seems that i will not have time to play with the `/fdm/jsbsim/propulsion/engine/AFR` prop for some time now. I propose we will merge this in the current state, as this is quite good already i think. I will open another issue (see #299) so we have a reminder.

hbeni:
I want to add the current status to the lever tooltip, so pilots can adjust to the right directoon with the cabin messages turned off.

hbeni:
nicer :)
should be really ready to merge now.

HHS81:
Will be merged!

HHS81:
Thanks a lot! 👍

PR #256, opened 2018-02-11
Added stall horn test and prepared icing

hbeni:
This adds:
- stall horn function dependency on active breaker and electrical power (see POH 7-47)
- stall horn test switch property to test the horn on the ground
- prepared stall horn icing (piggybag on pitot icing)
- stall horn can theoretically fail now (`/controls/stall-horn/serviceable` -> false)

Open:
- [x] add exterior clickspot (only in `.ac` file currently)
  - [x] verify blender model
- [x] add texture
- [x] add checklist item to preflight (left wing leading edge / POH 4-11) (was already there)

image

PR #298, opened 2018-09-15
Added enhanced documentation (fix #282)

### LINKED ISSUE #282: Enhance documentation
Recently many new features where developed, some of them complex and allowing fine tuning via properties. This is currently not documented well.

I want to add a more elaborate documentation file, describing the new systems and features more deeply.
What documentation format should we use? I propose markdown document as it displays nicely in github and is also easily human readable.

Ideas for contents:
- Describe failure system in more detail, especially
  - the random surprise modes
  - how they can be activated at startup from commandline/launcher
- Describe complex engine procedure options and their effects
- Random property interfaces (for example manipulated trough telnet or so)
  - manual friction allows reducing power
  - how to insta-kill engine
- State system
- Some FAQs on common situations and how to handle them (problems with starting, etc)

What would you add?

hbeni:
I wrote up a draft for enhanced documentation.
In my opinion it should be fairly complete already, however please read trough it to correct my bad english :)

The document refers to all currently implemented and proposed features, so it would be good IMHO to merge this once at least #266, #256 are merged.

Once merged, we should include a link to the Documentation.md from the wiki page, i think.


HHS81:
Merging make sense when the icing feature is complete. That`s the next thing I will finish

Issue #144, opened 2016-08-01
icing effect using multiple dirt texture


HHS81:
When icing effect implemented use multiple dirt map  in the shader for

HHS81:
What happend to this C182S?

- [ ] Obviously it flew through clouds of canned whipped cream
- [ ] Doc E. Brown used it - "_The way I see it, if you're gonna build a time machine into a aircraft, why not do it with some *style?*_ "
- [ ] attack of white flys

or

- [ ] the experimental icing effect?

![fgfs-screen-005](https://user-images.githubusercontent.com/12649375/33808780-33df355e-ddec-11e7-8371-4e03911bbfa9.jpg)

gilbertohasnofb:
This looks amazing!

hbeni:
Is it already influencing the flight dynamics and what can i do to de-ice? (Click "repair"?)
I cant wait to try the effect by myself. What causes icing?

HHS81:
Not yet

Currently by putting it into the sun (Flights into known icing condition are forbidden with this Type of aircraft.) I'm not sure if I should add ground deicing equipement, as small aircraft are not deiced like bigger aircraft (Flights into  known icing condition are not allowed).

Cause of icing: (simplified) Freezing temperatures and high humidity.

HHS81:
I use the already existing Code from the Seneca II,    works quite good, but need a bit to be extended to be more realistic.

hbeni:
It would be cool to have an option to manually remove frost and ice. The POH states this in the preflight i think.

wlbragg:
Really looks good. This would be a nice system for the heavy aircraft to have (when ready, someone might want to ping the forum with it). My thoughts on it for these light aircraft, if what you say is true about not allowed to fly into/during icing conditions, would be it is only useful for simulating a startup condition after sitting in freezing weather. Unless your interest in it is to model what would happen if you were to try to take off and fly in this condition. It certainly would add some realism to the conditions your trying to fly in.

wlbragg:
>My thoughts on it for these light aircraft, if what you say is true about not allowed to fly into/during icing conditions, would be it is only useful for simulating a startup condition after sitting in freezing weather. 

Then again we model frost and fog on the windshield when in theory we shouldn't be flying when that condition is possible, at least the frost portion.

HHS81:
It is a bit more complex. I will write later about.

HHS81:
@hbeni @gilbertohasnofb @wlbragg 
[...]

PR #308, opened 2018-12-09
Issue144

HHS81:
Fix of Issue #144  adding a more or less simple icing effect which has effects on lift, drag, weight, engine colling and propeller. It detects vertical layers of clouds, and icing effect increases with rain and snow. It is also visible on the exterior.


hbeni:
Very nice work and a great enhancement!

PR #309, opened 2018-12-11
KT76c multiplayer and enhancements

hbeni:
I enhanced the KT76c transponder:
- a little better performance demanding
- TST mode fully implemented
- "R" reply annunciator behaves correctly now (blink/steady/off, depending on mode and power)
- Multiplayer compatible to ATCPie


HHS81:
Great Work!

Issue #315, opened 2019-06-10
FGFS 2019  JSBSim: Unrecognized tag <switch> in the condition statement.


HHS81:
I got following error message and a segfault:
```
In file E:/AirCraftAddon/c182s/Systems/fuel.xml: line 71
Unrecognized tag <switch> in the condition statement.
```

HHS81:
It seems Stuart had already fixed it on the C182T branch

HHS81:
And it seem that the xml-parser don`t like special signs like `°` etc. anymore.....

PR #316, opened 2019-08-12
Fix for #315

hbeni:
Ported over from the c182t branch, still works on FGFS Version 2018.3.2 (i currently have no fresher install), can please someone verify?


HHS81:
Thanks, I remember that I wanted to fiy that, but than somehow I was interrupted... Thanks a lot!

PR #325, opened 2020-05-28
Refactored calculation of COM1/2 radio power-btn to electrical.xml

hbeni:
The reason was that the `power-btn` property was changed only by turning the knob.

However other scripts (like Red Griffin ATC) rely on it to get the "radio is turned on"-state.
The power-btn of com1/2 is now calculated and considered "on" when not only the knob is turned,
but also the radio is serviceable and has electrical power.

### DIFF
--- a/Models/Instruments/Avionics/kx165-1.xml
+++ b/Models/Instruments/Avionics/kx165-1.xml
@@ -690,6 +690,7 @@
                 <max>1.0</max>
                 <wrap>false</wrap>
             </binding>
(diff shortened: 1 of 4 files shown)


stuartbuchanan:
This looks fine, except you should just delete the code in kx165 rather than commenting it out.  That'll be much cleaner :).

If you do that I'll be happy to merge the commits.

Thanks!

hbeni:
> If you do that I'll be happy to merge the commits.

done :)

PR #327, opened 2020-06-01
FGCom integration with GMA 340 audio panel

hbeni:
Add integration of FGCom, so it is controllable from the GMA-340 audio panel and reacts to loss-of-power/failures


hbeni:
I tested this with two PCs and standalone fgcom instances. Worked good!

stuartbuchanan:
Looks fine from a code review perspective.  Thanks also for detailing the testing you've done - that gives me more confidence to simply merge this straight in.

Issue #141, opened 2016-08-01
Implement damage


gilbertohasnofb:
Create damage model:
- gear collapse
- wings light damage
- wings heavy damage (broken wings)
- engine crash (e.g. hitting the propeller in the floor)
- airplane crash

gilbertohasnofb:
@wlbragg would you be interested in this?

HHS81:
I'm not happy with the way the c172p it does. I always notice a hit on the framerates and a short hicup when a damage occours. 

My priority would be to make the aircraft fly like the real one - here making the engine let start and idle realistic. Then we should proceed to most common failures: tire blow, surface effects on braking,  front gear collapse,  engine damage,  prop strike,  icing. Flightgear is not a crash simulator. But things which will lead to a crash should be simulated.

gilbertohasnofb:
Yeah, I kind of agree with you. But right now if we hit the ground the aircraft will just bounce as if nothing happened. I think at least killing the engine or something like that should be implemented.

HHS81:
Depending on the impact angles and strength there are many things happening: Magnetos stops,  fuel lines breaks,  ... . Sometime a real aircraft just slides,  sometimes they break into thousand parts. It isn't simulated good enough anywhere. I don't like the behavior of the DR400 as  an example. I will think about how to solve this.

hbeni:
@HHS81 Do you already have an idea how this could be implemented?

HHS81:
Yes,  but: I have still issues open and since I don't live from my work here, there is no rush!

hbeni:
I was just asking because i may give you a hand here. Maybe i can start a new branch and work on the XML infrastructure and you add the modelling part later?

HHS81:
Sorry, but I want to have it different like the c172p. There are some things I don' t like it there, that's why it doesn't make sense to add any xml structure or anything like that right now. Thanks for your offer, but not needed right now here.

hbeni:
Okay! No Rush!

HHS81:
collection of ideas/ things to damage:

- [ ] gear collapse
- [ ] tyre blow
- [ ] wings light damage
- [ ] wings heavy damage (broken wings)
- [ ] flap overspeed (loss and jam)
- [ ] prop strike (using a more realistic method than c172p - already in the pipeline)
- [ ] fuel line damage due g-force (hbeni?)
- [ ] electric failure due to lightning strike
[...]

PR #328, opened 2020-06-02
Simple damage system

hbeni:
Implemented a simple damage detection System (as foundation to #141):
- [x] Basic system ported over from c172p and adjusted coefficients accordingly to already present values (`Systems/damage.xml`)
- [x] Added a simple damage handler (`Nasal/damage.nas`) that informs the pilot via text of broken things.
  Now we could practice smooth landings without killing the front gear :)
- [x] Added a simple gui to aircraft dialog showing all damaged parts, so pilot can check it.
- [x] Damage should be stored over sessions when initializing "saved" state. This should be an easy one, just had no time to test this yet.
- [x] It has <s>no</s> effect on flight dynamics <s>yet. This could easily achieved by integrating the damage values to the fcs calculations, </s>like it's done in the c172p
- [x] gear should collapse; again this could be made easily by adjusting the z-pos of the contact points

As said in #141, model/graphical stuff etc. is to be done separately; Heiko already has some ideas for that.

- [x] Port that over to the C182T
- [x] when wings break, fuel should be leaked and the sensor break: small damage should make a leak, fully broken should set fuel to 0 and fail the sensor.


HHS81:
@hbeni 
Sorry, I`m busy with real life stuff- no corona holidays for me.

So, you are now going the c172p way? I wasn`t happy with this, since I experience a significiant lower fps with the c172p compared with the c182s. 
Otherwise using the z-values of the contact points as suggested by you as damage indication (strength, affected systems) seems cheap.

hbeni:
Hi Heiko, i had the FPS in mind and that was the reason i implemented it using jsbsim, which should be quit fast. Its really just some really small calculations this far.
Also i really wanted to train my landings :)

I think the hiccups you experienced came from the effects with the 3d stuff. I could not trace performance issues here so far; do you have any now with the branch?

Also, this was not soooo much of work so far, so if you have a better way and time to implement it, we can just easily throw away this implementation. Most of it is in `damage.xml`.

stuartbuchanan:
Like Heiko, I'm not a huge fan of the c172p approach to adding very complex systems to what are fundamentally edge-cases and trying to model the impact of damage..

I think it's very good to notify the user when limits are exceeded that would damage the aircraft - wing G-loading, excess undercarriage loads, exceeding VNE.

However I'm less convinced about trying to _model_ that damage either graphically or in the FDM.  In my mind it is sufficient to simply provide a warning message to the screen.  Some users will ignore this, others will be trying to fly within the envelope.

In terms of damage modeling itself:
- Under excess loads, undercarriage on GA aircraft (microlights) tends to fail completely - the aircraft basically stops moving on the ground with a bent leg.  So there's little point in trying to model the effects of this - the aircraft simply cannot move under it's own power.

I also don't think modeling damage to aircraft after they exceed Vne is accurate.  As part of my annual check flight, my aircraft is flown as close to Vne as it can (e.g. steep dive, full power).  So I do not think it would be damaged as soon as it exceeds it in any real way.  

-Stuart

hbeni:
The problem is to detect that some value was exceeded. This is the part which was modelled in jsbsim, because it is simple and fast there.
Currently there is a text expression in red when this happens.

I agree that we really don't need some fancy 3D modelling of damage for now (however that would add the public value of the simulator in my eyes), but i think it is very useful that the plane will not be flyable anymore when the wings broke (they don't do it straight out and well above the Vne value currently), as well as a collapsed gear - it gives immediate visual feedback to the pilot.

hbeni:
No worries. I donÄt think it is done in stone, but i think what we have here is better than nothing. We may come back and improve anytime.

hbeni:
@HHS81 
> I like the way X-Plane handles it.

I don't know XPlane - how does it handle it?

HHS81:
@hbeni 
As I could see, similar now like you now implemented. They feature overspeed of flaps, tyre damage, and engine damage on harder landings. 
I will add some animations to the landing gears and see how I get the friction coefficient manipulated when gear is bent.

hbeni:
Animations would be nice but please stay sharp on performance...
I think the performance got in the 172 comes from switching the 3d models.

wlbragg:
>I think the performance got in the 172 comes from switching the 3d models.

That is right. It's almost entirely the extra mesh. The j3 was even worse as it had unnecessary 4kX4k textures. I think had I to do it all over from scratch, I would use shaders for the damage visuals. Something akin to wingflex.
As far as to do or not to do at all and peoples personal preferences. I think it is much more pleasing to see the results of messing up VS any message unless the message is "real time/real life" such as a stall warning siren or engine flame out. FDM  changes due to structural failures if done with much planning could really enhance the simulation for emergency landing training and the likes. I think visualizing gear damage or collapse should be fundamental to a good simulation. It was even crudely built into the core. The approach I took was before I had gained much experience with modeling or FDM components and could be vastly improved on. But I feel it is most certainly value added to the simulation in general.
But then again I also believe we should finish the ground cover effects on the wheels and gears in the core code. All of these things were obviously at least thought about originally, or we wouldn't have some of the structure supporting these things, like ground cover hardness and friction values.
All the code I added for the ground effects (sinking and friction) had little noticeable impact on performance. But it made the gear change outs useful for something other than esthetics. Now the bush gears work on terrain where the stock gear doesn't and the skies work on snow where the default and bush gear don't. I personally like having that ability and realism.

HHS81:
No fps impact noticed so far- but only prop and gear done yet.

crashed

hbeni:
Looks very good!
(Poor plane 😢)

wlbragg:
Looks good, if your into seeing your prop mangled, ha!

>That is right. It's almost entirely the extra mesh. The j3 was even worse as it had unnecessary 4kX4k textures. 

Now that I think about it, the first version of damage on the c172p was "more" expensive than the current version. The entire wing unit was one piece and required several duplicated of the entire wing bent in the different damage patterns. a set for left wing damage, right wing damage and both wings damage.
Similar to how the j3 still is. On top of this extra mesh were 4k textures.
cubdamage

I need to change the j3 to what we did to the c172p. 
c172pdamage

hbeni:
@HHS81 Heiko, what do you think: The damage inspection GUI was introduced from me because i had the impression, damage will not be reflected in the model in the near timeframe.
With your work, i think it's obsolete and can be removed (at least when the wings are bent) :)

wlbragg:
FYI: I noticed you may be using a "repairing" flag to use for settling damage when resetting damage, similar to what the c172p does. Note an unresolved issue with the repairing flag and timer on "Airport Relocation" using the GUI.
See https://github.com/c172p-team/c172p/issues/1325

hbeni:
> FYI: I noticed you may be using a "repairing" flag to use for settling damage when resetting damage, similar to what the c172p does. Note an unresolved issue with the repairing flag and timer on "Airport Relocation" using the GUI.
> See [c172p-team/c172p#1325](https://github.com/c172p-team/c172p/issues/1325)

Thanks for the heads up! i was not able to reproduce that with the c182s on this branch.

hbeni:
Oh god, it worked. Poor plane... Good that its so cheap to repair :)
I'm really thrilled how you will implement the broken wings 3D representation.

HHS81:
> 
> 
> Nice, it never occurred to me to do that. Good idea!

I didn`t like the fact that you can fly through trees of a forest and still get up into the air. I don´t want to have a crash simulator, but also not an arcade game, where you get signalized "game over".
Especially in Nort America with all those little airports hidden in the woods, the perfect place for a C182.

HHS81:
> 
> 
@hbeni 
> Oh god, it worked. Poor plane... Good that its so cheap to repair :)
> I'm really thrilled how you will implement the broken wings 3D representation.

First I want to find out, why the JSBsim structure points don´t recognize WOW. Then I will see...

hbeni:
Hey Heiko,

> First I want to find out, why the JSBsim structure points don´t recognize WOW. 


i tried to figure out what exactly you are after, can you elaborate a bit so i can assist?

stuartbuchanan:
Damage < 90ft AGL assumes that all trees are that high.

Looking through the regional material definitions, there are a lot of areas
where the height of trees is significantly lower than 90ft:

Here's a very rough distribution of heights:
10: 14
15: 24
20: 73
25: 63
30: 0

Obviously that is a count of regions, not the areas of the said regions.

I suggest lowering the limit to (say) 50ft (~15m).  That would ensure you
are hitting the tree below the tree-tops (we generate some variation in the
tree-top height as well), account for the c182 being high-wing, and the
wing being strong enough to work as a leaf-strimmer for small branches!

(Personally, I'm not a fan of complicated damage systems, and particularly
the effort to animate a crashed aircraft.  My experience IRL is that no-one
wants to look at a crashed aircraft except the accident investigator and
the engineer fixing it.)

-Stuart

hbeni:
Would it be possible to get the current regions average tree height (or even the tree height from the actual area) from the property tree? If so, we could easily make it dynamically.

HHS81:
> 
> 
> Would it be possible to get the current regions average tree height (or even the tree height from the actual area) from the property tree? If so, we could easily make it dynamically.

Unfortunately not. They are not shown in the property tree, otherwise I would like have done it that way. 😞

hbeni:
Should we open a FR ticket on fgfs core to get the property? :)

HHS81:
@stuartbuchanan 
> 
> Damage < 90ft AGL assumes that all trees are that high. Looking through the regional material definitions, there are a lot of areas where the height of trees is significantly lower than 90ft: Here's a very rough distribution of heights: 10: 14 15: 24 20: 73 25: 63 30: 0 Obviously that is a count of regions, not the areas of the said regions. I suggest lowering the limit to (say) 50ft (~15m). That would ensure you are hitting the tree below the tree-tops (we generate some variation in the tree-top height as well), account for the c182 being high-wing, and the wing being strong enough to work as a leaf-strimmer for small branches! (Personally, I'm not a fan of complicated damage systems, and particularly the effort to animate a crashed aircraft. My experience IRL is that no-one wants to look at a crashed aircraft except the accident investigator and the engineer fixing it.)
> […](#)
> -Stuart

Yes, that sounds like a good compromise. 
I don`t want to have crash simulator as well. On the othe side, we claim to be a simulator and not an arcade game with "game over". But I would like to have an effect when as an example exceeding rwys - flying through the trees is too simple. I don`t want to go further, because I think it is enough for the user to hear the bang hitting the trees and loosing altittude with ending sitting on the ground. It shows people that the ability of the aircraft is limited.

HHS81:
> 
> 
> Should we open a FR ticket on fgfs core to get the property? :)

You can do it if you want

HHS81:
> 
> 
> Hey Heiko,
> 
> > First I want to find out, why the JSBsim structure points don´t recognize WOW.
> 
> i tried to figure out what exactly you are after, can you elaborate a bit so i can assist?

I don´t get any WOW-signal of the structure points, when the gear is crashed

hbeni:
> > Should we open a FR ticket on fgfs core to get the property? :)
> 
> You can do it if you want

done: https://sourceforge.net/p/flightgear/codetickets/2389/

hbeni:
> Yes, that sounds like a good compromise.

I think that too. The crash-feedback is extremely valuable tough, especially with the gear collapse we can train good landings.

hbeni:
@HHS81 I tried it and can't reproduce (ie i get WOW contacts after crash) - is it this what you where looking after?

Screenshot_20201027_090139

hbeni:
So, i just ported missing bits to the C182T :)

@HHS81 Heiko, did my investigation above (https://github.com/HHS81/c182s/pull/328#issuecomment-717063039) help you already?

hbeni:
I just added a small enhancement: If the wings are damaged, fuel will leak out of the broken tank.
If the wing damage is severe (like ripping it off or collide with something), the whole fuel will be leaked instantly, and navlights/strobe will cease to work.

I think, the last thing missing here is the model showing the wing damage. the other stuff seems to work fine, both on the S and the T variant; i did plenty of test flights already.

HHS81:
> 
> 
> I just added a small enhancement: If the wings are damaged, fuel will leak out of the broken tank.
> If the wing damage is severe (like ripping it off or collide with something), the whole fuel will be leaked instantly, and navlights/strobe will cease to work.
> 
> I think, the last thing missing here is the model showing the wing damage. the other stuff seems to work fine, both on the S and the T variant; i did plenty of test flights already.

Hmm... the Cessna 182s has fuel bladders- I don`t think that fuel will leak with low wing damage.

hbeni:
> Hmm... the Cessna 182s has fuel bladders- I don`t think that fuel will leak with low wing damage.

Thanks for catching, i didn't know. The POH id not mention specifics :)
I fixed that.

Did my https://github.com/HHS81/c182s/pull/328#issuecomment-717063039 help you?

HHS81:
> 

> Did my [#328 (comment)](https://github.com/HHS81/c182s/pull/328#issuecomment-717063039) help you?

Not really- but I guess I do know why it behaves like that. I think we can merge

hbeni:
> I think we can merge

@HHS81 should i, or do you want to do it yourself?

PR #329, opened 2020-06-03
Davtron 803: Converted to canvas. This solves the "white text problem of OSGText on linux.

hbeni:


HHS81:
I added some further optimizations to the aircraft regarding Compositor rendering

PR #330, opened 2020-06-11
Adjusted davtron temperature display to round numbers (no fraxtions) as per manual

hbeni:
I rechecked the LCD display of the davtron and temperature should not read out fractions but just whole degrees.


hbeni:
Hello Heiko,
i think this is already fine and could be merged... OK if i do so?

PR #333, opened 2020-08-20
Small fixes for 2020.2

hbeni:
- Fix 2020.02 startup error complaining about missing `electrical/output/comm[n]` nodes
- Fix nasal magnetos selector error (was cosmetical)


hbeni:
Confirmed to work at the fgfs forum

PR #336, opened 2020-09-27
Upgraded KX165 to KX165A to support 8.33 channel spacing

hbeni:
Port the version from c172p over


hbeni:
Hi there,
i succeeded in porting the radio over. It currently shows and seem to work, based from the property tree. It responds to the failues and electrical system like the old one did.

However i have an issue i cant't wrap my head around - despite the radio is turned on, has power etc, there is nothing showed at the LCD.

It would be cool if @HHS81 or @gilbertohasnofb  or @wlbragg  could look over this in the `Issue331-UpgradeRadio`-branch, because i'm stuck and think this is only a minor thing.

As soon as this is fixed, i think we are probably ready to go...

wlbragg:
It looks like your using either or both
~~~
<property>/sim/model/c182s/lighting/comm0-power</property>
and
<factor-prop>/sim/model/c182s/lighting/comm0</factor-prop>
~~~
for power to the LCD, yet in the property tree those properties have no value. Maybe you missed a filter or logic statement?

wlbragg:
Yeah, look in c172p Systems/Instruments.xml for 
/sim/model/c172p/lighting/comm0-power (around line 230)

hbeni:
Thank you @wlbragg! That did the trick.

I fixed the lighting and integrated the glearshield lights. i think it looks good now!

=> Ready for test trough a second hand

hbeni:
I think the light effects are not entirely good:
grafik
- [ ] The digits have black background
- [ ] The cassis is too bright i think

Any idea whats wrong here?

hbeni:
(for the archives: just did a test against hbeni/fgcom-mumble#50 at LOWI and it did work like a charm :grin: )

hbeni:
Ok if i merge this?
@HHS81 

I think i tested enough...

hbeni:
@HHS81 ...maybe you can also have a look at this one?

HHS81:
> 
> 
> @HHS81 ...maybe you can also have a look at this one?

Do I need mumble?

hbeni:
No, just have a look at the model integration etc.
Properties changed to the ones from c172p and i tested that part already very toroughly.

HHS81:
@hbeni Hmmm... That how the real KX165A TSO looks like: https://www.pilotshop.com/catalog/avpages/kx165a.php
And that`s how ours look like: 
fgfs-20201026164953
I´ll check the reason for. And I´m not a fan of the uvmapping. The Bendix sign is hardly readayble, but the numbers are.
Btw.: On some pictures the led is red, on others it seems white/ greenish.

HHS81:
@hbeni simply wrong material settings ;-)

HHS81:
@hbeni Thanks for the work!

hbeni:
Thank you for checking!
Your thanks belong to the c172 project, i just ported their implementation

PR #338, opened 2020-09-29
Fix #337: Apply chocks only when their state differ 

### LINKED ISSUE #337: Got stuck in chocks
Today while testing i got an interesting behavior (FGFS 2020.3.0).
I did secure the plane after the last session and restored its state while starting using the overlay "cold&dark" state.
Testing showed the problem also occurs when applying "auto" state.

That resultet in:
- the aircraft securing dialog was showing chocks disabled
- chocks beeing on the ground, leaving me with an unrollable plane
- trying to add/remove chocks trough the GUI resulted in adding additional chocks. The original ones could not be removed.

hbeni:
(fixes getting stuck with overlay cold&dark)

### DIFF
--- a/Nasal/c182s-states.nas
+++ b/Nasal/c182s-states.nas
@@ -47,9 +47,9 @@ var secureAircraftOnGround = func(state) {
         return;
     }
     
-    setprop("/sim/chocks001/enable", state);
-    setprop("/sim/chocks002/enable", state);
-    setprop("/sim/chocks003/enable", state);
+    if (getprop("/sim/chocks001/enable") != state) setprop("/sim/chocks001/enable", state);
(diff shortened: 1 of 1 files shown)


hbeni:
Tested it, should be good for merge

HHS81:
Hmm... maybe the symptoms are solved. But not the cause. I experienced something often while develoing this feature. Reason alsways had been somewhere in the gui dialogs. I check again

HHS81:
Indeed, interesting. Only appears with cold and dark state pre-selection. Merged.

hbeni:
I think the actual problem was that the state nasal code always set the chocks property, and the listener to that adds a new model.
The fix was that the property is only changed if the value differs.

HHS81:
@hbeni yea, nice catch!

Issue #178, opened 2017-10-28
Implement Multiplayer compatibility


hbeni:
Make aircraft showing correctly in multiplayer with all liverys, esp. custom callsign one

- [x] xpdr has been merged:
  - Transponder does work in OpenRadar and ATC-Pie
  - Transponder IDENT does work too
- [x] Pending:
  - [x] show correct model
  - [x] Flight Control Surfaces
  - [x] different liveries
  - [x] registration in the custom livery (corebug -> see https://github.com/c172p-team/c172p/issues/1339)
  - [x] lights
    - [x] landing lights
    - [x] taxi lights
    - [x] beacon
    - [x] nav lights
  - [x] Cowl flaps

hbeni:
Just tested: base livery work already. Missing seems things like opening doors, landing lights, custom registration number showing, etc

HHS81:
Multiplay seems quite to need a lot of ressources. I do know a lot users who are complaining about other mp-aircraft, which slows down or even freezes the sim when they appear.

I don't think that the other user need to see the doors opening. What is important are lights, liveries and registration and flight control surfaces (incl. cowl flap).
All other this is not really needed.

hbeni:
Also, transponder working correctly with atc clients (ie openradar) should be verified. While beeing at this, we might make the factory vfr code setting accessible via the aircraft dialog.

HHS81:
@hbeni 
I can`t tell for sure, but I think the EC 135 P2s transponder supports the transmitting if the transponder code. So we might loook how it is done to complete the implementation.

hbeni:
I started something in the [`xpdrMP`](https://github.com/HHS81/c182s/tree/xpdrMP) branch.
ID Button is already working and the internals seem good - however no id is transmitted. I asked on the mailing list about this.

hbeni:
https://github.com/HHS81/c182s/tree/xpdrMP is now working in ATCPie :)
We can fly and interact correctly with ATCs now.

till open is the higlight of the Ident button, when the id is actively sendet.

hbeni:
To brush this up a little:
- [x] xpdr has been merged:
  - Transponder does work in OpenRadar and ATC-Pie
  - Transponder IDENT does work too
- [ ] Testing pending:
  - [ ] lights should be verified, but i expect them to work
  - [ ] different liveries
  - [ ] registration in the custom livery
  - [ ] Cowl flaps
  - [ ] Flight Control Surfaces

[...]

PR #339, opened 2020-10-21
MP fixes

Zaretto:
Fix model paths for MP

Fix JSBSim model min greater than max.


HHS81:
Hmmm... tested and I got this:

hbeni:
Heiko, exactly my problem too!
I’m on 2020.3, you?

HHS81:
The same. I got the correct model a while ago, but I have to test again, since I used a different FGFS-version then.

hbeni:
What also really worries me is my comment in #178 on 28 Oct 2017, where i stated that the base livery already works.
Now it does not anymore, what we see in your screenshot is the bare fallback-model! (when removing the `<fallback...>` node, you even get the blue/yellow glider, which tells me there is something off when fgfs tries to load the c182s model; however, that process works fine with the c172p!)

hbeni:
works again, see #178.
But there is still something weird happening here.

c182s.xml
```diff
@@ -758,8 +758,8 @@
                 </domain>
                 <range>
                     <!-- reducing factor: 1=no reduction/0=full blocked; min=no-ice, max=full-ice -->
```
hbeni, line 760: @Zaretto Are you sure this is correct?
The min/max output is meant to be a inflow reducing factor, thus, when ice is maximized, the reducing factor has to decrease to model "less air gets inside".

Ice >= 0.010 inches should yield in 0.75 reduction of air original air inflow.
With your fix, we get: Ice==0 yield 0.75 original inflow.

c182t.xml
```diff
@@ -449,8 +449,8 @@
                     <max>20</max>
                 </domain>
                 <range>
-                    <min>1</min>
-                    <max>-1</max>
+                    <min>-1</min>
+                    <max>1</max>
                 </range>
```
hbeni, line 454: @HHS81 I cannot see changes here.
What is affected by this property?

@Zaretto I think the original idea was to have the right aileron negative, when the left one is positive, and vice-versa.

hbeni (commented):
I left some comments in the code.

c182t.xml
HHS81: It is indeed like you said- right aileron negative etc... I borrowed the code from the c172p

Zaretto: my min/max changes were wrong- caused because I wasn't paying enough attention and I had clipto errors because of an old version of c182s.xml JSBSim on the path.

Now reverted; the rest of the pull request is still valid and should be reviwed.

Sorry for the noise.

hbeni:
Hm, still happening when the model is not loaded:
grafik

Otherwise it looks good

Zaretto:
With my changes I don't see the interior any more.

image

I've made a change to the fgdata fallback model to use the c182s-dpm rather than c182s-ai because the -dpm model is almost in the same position as the high detail model. I'll also put this change into the 2020.3 branch

hbeni:
> With my changes I don't see the interior any more.

I was testing commit e5b5fce7b5 on FGFS commit  b6bcaaa36d5d65

hbeni:
All right. I retested with significant altered visbility settings, there it works.
The default is way too conservative i think and also causes the cockpit showing behaviour; i think this is the mismatch when the lowres fallback model is displayed and at the same time the interiour shows, because of the model displacement you already mentioned.
I obviously didn't got the latest change to fgdata to get the c182s-dpm.

So, from my perspective this all looks good now! :+1: 
Good work, and i will definitely want to base further MP work (more properties) on that one.

@Heiko: are you fine with merging now?

hbeni (changes requested):

c182t.xml
```diff
@@ -449,8 +449,8 @@
                     <max>20</max>
                 </domain>
                 <range>
-                    <min>1</min>
-                    <max>-1</max>
+                    <min>-1</min>
+                    <max>1</max>
```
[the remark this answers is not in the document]
Zaretto: I managed to got the revert wrong; should now be fixed.

hbeni:
looking good now i think

hbeni (approved):

HHS81:
> @heiko: are you fine with merging now?

I will check this evening

hbeni:
@HHS81 Heiko thanks for testing this so fast, this will allow me to complete the missing animations in #178

HHS81:
> 
> 
> @HHS81 Heiko thanks for testing this so fast, this will allow me to complete the missing animations in #178

So much as I could test (I only have one computer and often it segfaults when I start the other instance of FlightGear only see my own aircraft. I tested an aircraft without those fancy mp-animation and I can see the other aircraft

HHS81:
@hbeni @Zaretto 
I got it- really need the command line when you set up two instances of FGFS on one computer. 
That´s how it looks like:
Unbenannt3
The registration isn`t transmitted yet, and the wheel fairings. Not sure if we need the wheel fairings.

Btw.. I´m surprised that the latency is quite big: 
Unbenannt2

HHS81:
@Zaretto github gives me conflict warning of Models/interior-model.xml.... any idea?

hbeni:
> The registration isn`t transmitted yet, and the wheel fairings. Not sure if we need the wheel fairings.

Yes, i will adress that when tackling #178 once this is merged.

hbeni:
>@Zaretto github gives me conflict warning of Models/interior-model.xml.... any idea?

@HHS81 i did a local rebase and it merged fine into master (locally), so i think, Richard just need to `git fetch --all; git rebase upstream/master`

Zaretto:
I merged in master and that has fixed the conflict

hbeni:
Today i tested this with the ORCAM, but there it looks like always the default livery is shown. Also, no lights show...
Is this a ORCAM bug, or still a problem with our model?

When i do a local-lan MP test instance, the livery updates a few seconds after launch.
Lights are shared among both instances, but that is surely another absolute-property-location issue, like pitot-cover and winter-kit, and surely some other, than can be dealt once this is merged.

```
#--multiplay=out,10,127.0.0.1,5000 --multiplay=in,10,127.0.0.1,5001 --callsign=Test1
#--multiplay=out,10,127.0.0.1,5001 --multiplay=in,10,127.0.0.1,5000 --callsign=Test2
```

hbeni:
Fixed that one above: my problem was my local setup:
`50.14 [ALRT]:io         loadxml: reading 'Path "/home/beni/Dokumente/FGFS/Aircraft/c182s/Models/Liveries/D-ELFP.xml"' denied (unauthorized directory - authorization no longer follows symlinks; to authorize reading additional directories, pass them to --allow-nasal-read)`

So, finally everything seems ready for merge to master, so i can continue working out the other properties.

hbeni:
> So, finally everything seems ready for merge to master, so i can continue working out the other properties.

I have a local branch ready that makes all the other MP stuff work. See also #178.

hbeni:
I close this one, the changes are now part of PR #351

Issue #341, opened 2020-10-28
Integrate new default descriptive Properties


hbeni:
Just a reminder:
There are  some new properties that will allow the GUI to make better decisions:

http://wiki.flightgear.org/Aircraft-set.xml#Performance_and_Flight-planning

PR #342, opened 2020-10-28
Implement #341(Integrate new default descriptive Properties)

hbeni:
Additional descriptive tags in aicraft-set.xml


hbeni:
XML is sanitized and loads (here) without visible errors.

hbeni:
Just another comment:
- i have no clue how to actually test this
- i chosed to use bare minimum POH values for the minimum runway length. Those are only achievable at sea level using stol techniques and thus not in all cases.
Should we maybe use more conventional values here? From the wiki i learned that the numbers may be used by the launcher to issue a warning and the map to hide shorter airfields...

HHS81:
> 
> 
> Just another comment:
> 
>     * i have no clue how to actually test this
> 
>     * i chosed to use bare minimum POH values for the minimum runway length. Those are only achievable at sea level using stol techniques and thus not in all cases.
>       Should we maybe use more conventional values here? From the wiki i learned that the numbers may be used by the launcher to issue a warning and the map to hide shorter airfields...

I would choose the bare minimum POH - if users uses the aircraft in a realistic way, they would have made a flightplan, calculated fuel, calculated W&B, checked weather, and decided if the airport strip is long enough.

hbeni:
Then we can merge 😁

PR #344, opened 2020-10-30
Issue 343 flaps operation + lights consume power

hbeni:
Operating the flap actuators as well as turning on radio lights or glareshield consume electrical power, showing on the ammeter


hbeni:
Tested, looks good for me.

PR #346, opened 2020-11-02
Issue 345: Convert splash screen to new format

hbeni:
Added Aircraft meta info for the C182S:
- Converted splash screens to new previews system
- added Authors, Maintainer and URL, so the information gets displayed by the QT launcher
- Made new cockpit shots


hbeni:
@stuartbuchanan should we add author/url/maintainer to the C182T variant too?

hbeni:
> @stuartbuchanan should we add author/url/maintainer to the C182T variant too?

I did it, maybe you can have a look. but i think it looks good.

hbeni:
I think this is merge-ready too.

PR #347, opened 2020-11-03
Issue 193: Implement alternate static air

### LINKED ISSUE #193: Implement alternate static air
see http://wiki.flightgear.org/FGproperties/Systems/static

hbeni:
Fix #193 for  C182S and C182T:
- Adds a second static source (primary is on both fuselage sides, second is in cockpit, POH 7-42; note that the cabin pressure is not as accurate than the fuselage static ports)
- Makes the source selectable with the alt-static cockpit knob
- Selected source is supplying the instruments
- Both static sources can fail (set `/systems/static[n]/serviceable=0`). The failure manager can fail the first port from the GUI.
- The fuselage static ports are subject to icing


hbeni:
Tested positively with both models.
- Select CAT-III minimums, cold OAT and fly
- Watch the icing accumulate (`icing` and `serviceable` properties of `/systems/pitot/` and `/systems/static/`)
  - first blocking the pitot (which can be remedied by heating it)
  - and after a while blocking the fuselage static port (which can be remedied pulling the alt static knob)

wlbragg:
Ooh, can I steal this for the c172p? We have the switch but not the logic?

HHS81:
> 
> 
> Tested positively with both models.
> 
>     * Select CAT-III minimums, cold OAT and fly
> 
>     * Watch the icing accumulate (`icing` and `serviceable` properties of `/systems/pitot/` and `/systems/static/`)
>       
>       * first blocking the pitot (which can be remedied by heating it)
>       * and after a while blocking the fuselage static port (which can be remedied pulling the alt static knob)

👍

hbeni:
> Ooh, can I steal this for the c172p? We have the switch but not the logic?

No. I just have learned about the venturi effect and i cannot lie, i must see this implemented before you can copy it 😁
That is a minor thing systems wise, but i really want to try to look up good strong that effect is.

hbeni:
@wlbragg Night shift done: Venturi effect is here, and an bug with the autopilot was fixed.
The effect is not _perfectly_ matched to the POH but reasonably close (and the numbers in the POH are rounded anyways!)

hbeni:
Actually the autopilot stuff should maybe be rechecked:
- Currently it is plumbed to the aircraft static system and thus subject to the valve pulling.
(previously it was plumbed to the static[0] source)
- I'm not sure if this is right: https://bkx.bendixking.com/downloads/006-18034-0000_3.pdf says (i interpret it this way) that it may have its own static source.
- I was not able to find more information about that,but the manual leads me to the conclusion, that the KAP140 has a individual static port on his chassis, and therefore should not be subject to the static system of the plane, but read cabin pressure values.

If that is the case the fix would be easy:
- revert 97307c61d77158b2
- add another static source with the Venturi effect applied (ie. in the cockpit) exclusively for the autpilot, that cannot fail and is not subject to the other static ports
- adjust the Line https://github.com/HHS81/c182s/blob/5e90351a7220d17396ceb9f6947d7d4f6fe24f40/c182s-set.xml#L1117 to that property

But in my opinion that would make no sense at all - that would mean that the autopilot always would fly an "off" altitude in regards what was entered in the altimeter and the autopilots baro setting.
Or we just plumb it to the outside atmosphere pressure property (but that would also be wrong...)?

What are your Opinions on that, @HHS81 and @wlbragg?

... and while we are at it: Opening windows has an effect on cabin pressure too, but it's not simulated so far. I even do not know which numbers i should use, and the effect will be dim, i think, so i stay away from that hassle, OK?

----
Excerpt from the manual:
grafik

wlbragg:
Static pressure in the KAP140
>What are your Opinions on that

Yeah, that is what it says, it has its own sensor.

hbeni:
Thanks for commenting. I tried to set it to the cockpit pressure, but that gives weird autopilot behaviour, especially after takeoff (engage the AP in climb and it will nosedive). It seems that the generic KAP autopilot simulation doesn't handle that well enough.
=> I vote for keeping it tied to the selected static source for now.

wlbragg:
If the interior sensor is messing it up, then I agree. Don't know if the KAP-140 can use a different (external) sensor though.

hbeni:
> Don't know if the KAP-140 can use a different (external) sensor though.

It can, but as soon as the source of the altimeter and KAP disagree, it messes up (like when climbing/descending)

hbeni:
I think we should leave the static port of the autopilot where it is now, as i think the KAP can’t simulate it well yet.

Otherwise my tests were ok, so from my point of view we should merge this feature.

hbeni:
For the archives: A new info re which static port to use clarified the KAP connects to the planes static system (instead of providing it's own):
https://github.com/HHS81/c182s/pull/379#issuecomment-791627709

Issue #349, opened 2020-11-09
ASI stuck at 160 KIAs


hbeni:
In current master, the ASI stops right at 160 knots indicated airspeed:
![grafik](https://user-images.githubusercontent.com/13608602/98565237-a8b43100-22ad-11eb-9386-e2a85d6ae0c9.png)

hbeni:
Reason: The `Airspeed.xml` uses the property `fdm/jsbsim/velocities/vias-kts` which is stuck at 160.
It is defined in `Systems/indicated-airspeed.xml`.
The reason is that the table is not interpolated outside the given parameters.
Fix is therefore to add such a mapping.

PR #352, opened 2020-11-21
New correct Airspeed indicator model (Fix #159)

### LINKED ISSUE #159: Wrong Airspeed Indicator model
The knob of the airspeed indicator is currently on the right, but should be left:
image

hbeni:
New Airpseed indicator from cessna, modeled after real Photo i took myself

Fix #159
Related to PR #160, #349

asi_new3
asi_new4
asi_new2
asi_new1

Texture is added to the Texture repo (see HHS81/c182-Textures#5)


hbeni:
Automatic on the C182T too :)
grafik

HHS81:
👍

HHS81:
Hmm.... I´m just testing it. 
I see some kts faster what fgfs or jsbsim provides
fgfs-20201121163148

HHS81:
@hbeni it seems it shows always 2ktn more, I just tried with different speeds, altitudes and temperatures. Othewise it looks geat!

hbeni:
Against which property do you check?
The ASI uses some jsbsim value not the one from /instrumentation... the comment in the xml said that this is on purpose because there is some error calculated from the poh in that value

hbeni:
Ah, had to zum the screenshot (I’m at my mobile right now).
You refer the TAS.
That’s odd, why didn’t i seen this?
I will try to update the texture later.

hbeni:
@HHS81  just to be sure, your altimeter was set to std pressure, right?
Your screenshot shows 29.98 instead of 29.92, which of course would yield a wrong readout...

HHS81:
> 
> 
> @HHS81 just to be sure, your altimeter was set to std pressure, right?
> Your screenshot shows 29.98 instead of 29.92, which of course would yield a wrong readout...

I set it to the value the ATIS gave me

hbeni:
Then it’s ok: the TAS OAT dialin value in the upper window is calibrated against pressure altitude (it also says that on the gauge), so you need to get your pressure altitude first by setting altimeter to std pressure.
Then dial in the oat, revert your altimeter to the atis value and you should get current values.

hbeni:
@HHS81 Cool that you like it :) I'm a little proud of that gauge. Was alot of work, but it turned out good i think. The SVG in the textures repo is dynamically, so it might server for other arcs as well :)
Making the TAS-disk was also not as complicated as i initially tought, it was just much fiddling around. I did alot of rotating in inkscape.
Other than that, calibrating it was not that hard: The 1000ft setting is the baseline which equals the IAS readings. The scale for the upper settings can be worked out from the upper Window markings by rotating and marking.

HHS81:
@hbeni 
I followed the steps mentioned in this tutorial: https://www.youtube.com/watch?v=kPpti6-jHaQ
Quite fast and easy.

PR #353, opened 2020-11-27
Update C182T

### LINKED ISSUE #335: C182T: Interior Model fixes
Hi @stuartbuchanan,
i think the c182T cockpit could be more user friendly. Currently only the switches body is clickable, which makes it hard to hit in turbolences etc.
In the c182s Heiko and i added also the base of the switch as clickable, so its way more easy to hit the switch.

- [x] I think that clickspots for the switches would also be a minor addition that would greatly enhance user experience.

-----
- [x] I just noted that the UV-Map of the Fuel selector is slightly off, so halve of the lever is black.

-----
- [x] Also noted that the Avionics breakers are not labelled. The labels are in the panel PNG file, however.

### LINKED ISSUE #348: C182T: Port systems from C182S
The T variant still misses some recent additions that may be ported over:
- [x] complex engine procedures / fuel system (partly working already? needs to be rechecked)
- [x] fog and frost
- [x] external heater
- [x] failure manager (failures do partly work, but needs adjustments for T variant which is missing some and adding others)
- [x] Air influencing the doors/windows/drag

hbeni:
I have noticed that some of the EIS display values are not correctly displayed.

This PR fixes this by loading an adapted EISController and publisher.

That now:
- enables correct priming procedure (fuel flow for priming was not displayed before)
- shows oil pressure and oil temp effects in bad oil conditions

-----
Also i noticed that the T variant is not up to date compared to the S. The PR also introduces all recent systems and features to the T variant (icing, complex engine procedures, door/window drag, failures, ...)

----
Fix #348

---
fixes from PR #365 
- breakers panel: texts added
- Fuel selector UV mapping corrected
- clickspots for the switches added
- fix #335

---
unique additions
- refactored failure manager to be able to distinguish the two plane models
- Propeller Governor can fail now
- Propeller Governor can't keep up if not enough oil pressure (fail-forward)


hbeni:
I tested this, so should be mergeable.

hbeni:
Added the last few bits; so we should be good for merging here now :)

hbeni:
Should be ready to be merged :)

HHS81:
There are conflicts in the failure management... ?

hbeni:
I already know these, no worry. Will fix them soon!

hbeni:
> There are conflicts in the failure management... ?

Resolved. 
Reason for divergence was that this PR for the C182T separates the failure code into the two planes;
but meanwhile the new KAP140 was implemented for which i added the failure modes in the legacy code already.

hbeni:
looked good when I tested it just now.

PR #362, opened 2020-12-15
Plane shakes a little when engine ignites

hbeni:
This adds a little eye candy effect: the plane shakes a little when the engine fires.


HHS81:
Nie one! 👍

PR #365, opened 2020-12-17
C182T interior fixes

### LINKED ISSUE #335: C182T: Interior Model fixes

hbeni:
- breakers panel: texts added
- Fuel selector UV mapping corrected
- clickspots for the switches added

fix #335


hbeni:
I just merged this into the `c182t-portSystems` branch (PR: #353) for easier final merge

PR #370, opened 2020-12-20
KAP-140: Fixes

### LINKED ISSUE #369: KAP-140: REV mode broken
The REV mode does currently not do what is expected from the Bendix manual.

KAP-140 REV mode did not work. The manual says, we need to dial in the FRONT INBOUND course to the HDG-Bug (and also OBS on VORs)

Details/tests in the C172p Issue https://github.com/c172p-team/c172p/issues/1302
Fix is simply to use the back course of the HDG-Bug on REV-ARM and REV modes.

hbeni:
- [x] Fix #369: REV mode did not work: The manual says, we need to dial in the FRONT INBOUND course to the HDG-Bug (and also OBS on VORs)
- [x] ALT ARM bug: ALT-ARM is possible with disabled AP (-> is a core isse https://sourceforge.net/p/flightgear/codetickets/2489/)
- [x] Enable switching between modes NAV/APR/REV (-> is a core issue https://sourceforge.net/p/flightgear/codetickets/2488/)
<s>- [ ] Enforce BARO setting after startup sequence</s>
<s>- [ ] (maybe: implement startup sequence after PWR-ON? we could do that pretty simple probably with an Nasal/Canvas combo)</s>
<s> -[ ] fix oszillation when heading INBOUND to a VOR </s>(see: #372)

### DIFF
--- a/Systems/KAP140.xml
+++ b/Systems/KAP140.xml
@@ -8,9 +8,57 @@
 <PropertyList>
 
     <filter>
+        <name>heading bug backcourse</name>
+        <debug>false</debug>
+        <type>gain</type>
+        <input>
+            <condition>
+                <less-than>
+                    <property>autopilot/settings/heading-bug-deg</property>
+                    <value>180</value>
+                </less-than>
+            </condition>
+            <expression>
+                <sum>
+                    <property>autopilot/settings/heading-bug-deg</property>
+                    <value>180</value>
+                </sum>
+            </expression>
+        </input>
+        <input>
+            <expression>
+                <sum>
+                    <value>-180</value>
+                    <property>autopilot/settings/heading-bug-deg</property>
+                </sum>
+            </expression>
+        </input>
+        <output>
+            <property>autopilot/settings/heading-bug-deg-backcourse</property>
+        </output>
+    </filter>
+    
+    <filter>
         <name>heading bug error computer/normalizer</name>
         <debug>false</debug>
         <type>gain</type>
+        
+        <!-- REV mode -->
+        <input>
+            <condition>
+                <equals>
+                    <property>/autopilot/KAP140/locks/rev-hold</property>
+                    <value type="bool">true</value>
+                </equals>
+            </condition>
+            <property>autopilot/settings/heading-bug-deg-backcourse</property>
+            <offset>
+                <property>instrumentation/heading-indicator/indicated-heading-deg</property>
+                <scale>-1.0</scale>
+            </offset>
+        </input>
+        
+        <!-- Other modes -->
         <input>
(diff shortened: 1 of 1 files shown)


hbeni:
@HHS81 I want to port over some fixes from the 172 KAP implementation that came to light when i tested this.

HHS81:
> 
> 
> @HHS81 I want to port over some fixes from the 172 KAP implementation that came to light when i tested this.

Sure. I added another bug: dial in any VOR; set Radial and HEADING towards. The aircraft will oszillating towards the VOR, not following the radial. After flying over the aircraft will finally follow the radial.

hbeni:
> I added another bug: dial in any VOR; set Radial and HEADING towards. The aircraft will oszillating towards the VOR, not following the radial. After flying over the aircraft will finally follow the radial.

I know what you mean, i think this is a problem with our PID controllers. They always overshoot the radial and need time to center on. I hope this will get fixed with the new implementation that is in the works over at the c172p, as tuning that is over my skill (i already spent many hours trying to do so :cry: ). But i have hopes: the new implementation has some property that tells us the time-to-interception and that probably is already used for that kind of thing, i hope that will solve it. But for that, the PIDs of that new implementation need to be adjusted first....

hbeni:
Heiko, can we split off your bug addition to another issue?
I feel it will take a long time to really fix finally and thus it will block the pull request...

hbeni:
Heiko, second:
I really think, that the remaining two cosmetic issues from me (enforce baro setting at startup and show testing stages at atartup sequence) need not be done now. The reason is, they are already modelled in the new KAP140 from Sascha, and that really would duplicate work. So maybe we should just hunt bugs here in this field until we can adopt the new, fancier one.

And regarding the oscillation: I would again propose that we open a new ticket for that here to not forget it.
and then we may close this one here and merge it, so we are basicly ready for a FGAddon release (that would be awesome)

hbeni:
Thinking about (and just played with it) i think the inbund oscillation is not a problem with the pid, but rather one of the input properties.
I still have no clue what I’m talking about, but what really puzzles me is that it’s just on one side of the VOR.
And rev mode is also affected.

maybe it’s just a sign reversal or wrong order of arguments to a calculation somewhere?

hbeni:
> maybe it’s just a sign reversal or wrong order of arguments to a calculation somewhere?

Played with the autopilot for about 2 hours, could not get anything to work. :-1: 
i think we will have to live with that for now...

hbeni:
Did open #372 for you.

I think we are ready to merge this now.

HHS81:
> 
> 
> Did open #372 for you.
> 
> I think we are ready to merge this now.

I merged it and will add soon a first fix for the oscillation

PR #377, opened 2021-01-23
Made Autopilot way more stable in HDG mode

hbeni:
Hi Heiko, today i played a little with the dual control and tested autopilot along the way.
In certain load situations (slow speed/heavy load; but especially speed) the autopilot started to oscillate in HDG mode. This happened way above the minimums of 80 knots.

I digged into this and remodeled the HDG mode:
- The HDG mode now does not directly try to match the heading bug difference to zero; instead it just calculates a command turn rate to reach that heading.
- HDG now rides ontop of the ROL mode, so we remain with just one PID controller, that was already tuned good.

The result is a way more stable and also smoother HDG mode, especially in the circumstances just laid out, but also on normal cruise speeds and also lighter loads.

I tested this for several hours now, and I'm quite satisfied.
However, as thats just not a minor change, it would be good if someone else has an eye on it before merging.

@HHS81 @wlbragg


HHS81:
@hbeni 
From my understanding it seems even that this is the way the real KAP140 is working. I guess the same way is needed for VOR/LOC mode.
I will test this evening, I´m pretty busy with studies right now

hbeni:
> I guess the same way is needed for VOR/LOC mode.

Yes, already accounted for: NAV/APR/REV calculate the target-intercept-angle for the current vor-needle deflection; and this is then used by the hdg mode (which is active for nav/apr/rev internally) to compile the needed bank angle, which is then finally fed into the rol-axis-hold.

Basically speaking: the AP has now just one lateral hold mode internally which is always on and bound to the turn-rate-indicator. The angle to hold is an input which is calculated from various sources (`0` for ROL, hdg-bug in HDG, or cdi-needles for NAV/etc).

i actually flew all the basic operations yesterday for testing (all modes worked as expected yesterday)

HHS81:
@hbeni I tried it now- and it works really great! I think we can change the servo rate now to really smooth

hbeni:
eb08cbc: Oh oh....:

grafik


Thank god, i made it. 10kn crosswind straight from the left...
grafik

(the random failure surprise modes can fail the servos too :grin: )

hbeni:
> @hbeni I tried it now- and it works really great! I think we can change the servo rate now to really smooth

Cool! I already made the roll servo smoother and i think its good, however, as i have no clue of real life servo speeds, you are free to experiment with the rate. But be sure to also test at about 80knots at max weight (actually i tried to make it stable at least for 75knots and overweight, to have some wiggling room).

I also do think that our autopilot XML is a great foundation for the new KAP model sascha has in the works. I just need to figure out how to wire it, and wlbragg already said, thats not easy.

wlbragg:
If there is anything I can do to help, let me know. I would need to take another look at it and compare the differences, maybe they aren't as much as I think?

I need to replace our original KAP with yours and then do a comparison with sascha's.

To get the most complete version of your KAP what branch should I pull from?

hbeni:
I added you modify to get you informed, since the last time i checked the new kap model in the 172 was as unstable as here at times.

The most current version is here: `AP-fix-HDG`

hbeni:
I tune it according to this: https://www.youtube.com/watch?v=MLNLZqaT9fI

hbeni:
Done, do you want to have a look?

If you like it, just merge :)

hbeni:
I integrated that fix in #379, so closing this as we are really close with that new KAP.

PR #381, opened 2021-01-28
Add turn-indicator inop flag

hbeni:
Quick lunch break start:

grafik grafik
grafik grafik


hbeni:
While testing i found a minor problem in the elec system.
The electric instruments should be wired to the ignition switch, and that should only output power when the INST-Breaker is in _and_ the key is turned away from `off`.

grafik

Will fix that later in this branch.

hbeni:
OK turned out, it's ok already:
grafik

@HHS81 so i will merge - if you agree?

HHS81:
Nice work!

HHS81:
I´ll merge it, as there will be another important fix come from me after that

PR #384, opened 2021-02-05
Issue383: Yaw due to aileron?

HHS81:
Fixed Yaw due aileron behavior - (probably wrong foresign in the data source); changed some parametres due longer ago JSBSim changes; made gears more springy as seen on real aircraft; synchronized c182t


HHS81:
Would be cool, if this could be merged soon, so the KAP140 can be tuned to it

hbeni:
Why are there so many commits listed above?
I assume we don't want all of them merged to master yet?

HHS81:
> 
> 
> Why are there so many commits listed above?
> I assume we don't want all of them merged to master yet?

Oh. damn, I think I have merged master into some while ago.... I will make a new pull request

HHS81:
faulty push request- new one on the way

PR #386, opened 2021-02-06
New Livery: G-XLTG

HHS81:
Signed-off-by: hhs81 <heiko.h.schulz@gmx.net>


hbeni:
Nice one!

💡 just an idea: the new canvas Registration allows for changing fonts and colours - we may leverage that to allow for example this livery (and the default one, etc) to have dynamic Registration.
What do you think about that?

@HHS81

HHS81:
@hbeni 
> 
> 
> Nice one!
> 
> 💡 just an idea: the new canvas Registration allows for changing fonts and colours - we may leverage that to allow for example this livery (and the default one, etc) to have dynamic Registration.
> What do you think about that?
> 
> @HHS81

This livery is based on a real livery (https://abpic.co.uk/pictures/registration/G-XLTG). There aren`t so much interesting liveries for Cessna 182S (the T-variant has more, and the older ones anyway). So for real based liveries I woulnd`t like to see dynamic registrations.
But I may produce some Retro-style liveries, where we could use this.

hbeni:
Yes, and the boring-grey default Liberty with c182 as Registration may also be dynamic (with c182 as default?)

HHS81:
Yep, that would be the first candiate! 👍

PR #391, opened 2021-03-05
Fix #390 "KT-70 legacy instrument compatibility" warning

### LINKED ISSUE #390: "KT-70 legacy instrument compatibility" warning
On next i see this on the terminal:
`5.00 [ALRT]:instruments KT-70 legacy instrument compatibility. Please update aircraft to use transponder directly`

hbeni:


hbeni:
Just tested it against ATC-Pie: Transponder still works as it should in MP (and in the cockpit of course)

PR #392, opened 2021-03-06
Optional transparent yoke-hiding for c182s/t

hbeni:
Over the past years I had the issue (here and there) that i had the yoke hidden to be able to read the instruments/knobs for takeoff and sometimes I was not centered by accident which gave bad takeoffs.

Today this idea struck my mind and I think it improves the user experience alot, since now we are able to see the stuff behind the yoke _and_ being able to see where the yoke is :grin: 

Users might tune the transparency setting to their liking (between fully-transparent to fully-opaque), and the setting is saved and restored across sessions.

----

grafik

grafik

grafik

grafik

### DIFF
--- a/Models/interior-model-c182s.xml
+++ b/Models/interior-model-c182s.xml
@@ -764,15 +764,13 @@
         </hovered>
     </animation>
     
- <animation>
-        <type>select</type>
+    <animation>
+        <type>material</type>
         <object-name>Yoke.L</object-name>
         <object-name>Yoke.R</object-name>
-        <condition>
-            <not>
-                <property>sim/model/hide-yoke</property>
-            </not>
-        </condition>
+        <transparency>
+            <alpha-prop>sim/model/hide-yoke-alpha</alpha-prop>
+        </transparency>
     </animation>
     
 <animation>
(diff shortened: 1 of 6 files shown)


hbeni:
@wlbragg this may be a quick user experience enhancement for the c172 too?

wlbragg:
I like that idea a a lot. I'm wondering if there would be any need for completely removing it anymore if it were to be made transparent. I think somewhere between your two different transparencies would be about right. Although I am only going by the pictures. It might be different on my monitor and in sim.
I'm always hesitant to add choices that do essentially the same thing, ie, remove or transparent. Transparent is better as it still retains the visual reference you need and yet allows access to the controls on the panel. So I think I would change the hidden option to only transparent. Great idea!

hbeni:
The way it’s implemented is just an on/off option; and when „on“ the transparency is applied. The range for the ammount is selecteable by the user (if he wants to change the three transparent default).

...or do i misunderstand your statement?

wlbragg:
No, I misunderstood. I didn't notice the slider. Perfect, great implementation.

HHS81:
Yep, that´s a cool  idea!  I like it!

PR #400, opened 2021-03-17
Fix Issues 287 399 (state system fixes/additions)

### LINKED ISSUE #287: State support using wrong launcher property name
I was looking at you code for ideas for the c172p state support system and noticed you are reading the property
/sim/init-state
in c182s-states.nas

I don't see that property being passed at all from the launcher. What I do see passed from the state flag is /sim/aircraft-state

Maybe it has changed since you set this up, or I am missing something about your nasal code.

### LINKED ISSUE #399: Aircraft options dialog: change state to drop down
The current state radio buttons need alot of space.
It may be better to switch to a drop-down instead.

hbeni:
- Fix #287
- Fix #399
- take-off and cruising state now fast-boot the new KAP140 autopilot
- "saved" now has a better viewing display in launcher

### DIFF
--- a/Nasal/c182s-states.nas
+++ b/Nasal/c182s-states.nas
@@ -15,9 +15,8 @@
 
 
 # The supported states in state property
-var supported_states = ["auto", "saved", "cold-and-dark", "ready-for-takeoff", "cruising"];
 var state_property   = "/sim/start-state"; #saved state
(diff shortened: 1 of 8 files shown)


hbeni:
Tested and good to merge in my opinion! :+1:

PR #401, opened 2021-03-25
Davtron 803M: Add countdown mode and timer alerts

hbeni:
This converts the davtron to a mostly property-tree-based implementation, so the nasal part and load gets easier. it particularly does just update the canvas on-demand now.

Furthermore, and more importantly it adds the following cool features; now fully behaving like the manual describes:
Manual link: https://www.davtron.com/cmsAdmin/uploads/M803_Front_Brochure-final-12-13-11.pdf (in particular page 2)
- The flight timer now always shows HH:MM format (previously it did show MM:SS until the wrap)
- The mode annunciator will flash normally, unless the device is in "setting mode".
- Pressing both SEL+CTL buttons together (_shift click_ one of the buttons) will enter the "setting mode" for the FT and ET modes (setting the clock was leaved out purposely because those are synchronized to internal time properties)
  - in FT mode, you can now set a flight time alert: Just press CTL button to change the blinking digit, and SEL to change the selected position. Once the fourth digit was confirmed by pressing SEL, the timer alert gets armed (if the entered time is in the future). Once the flight time is reached, the clock flashes and an aural annunciator plays some music. Confirm/quit the alert by pressing either SEL or CTL.
  - in ET mode, you can set a countdown alert in the same manner. Once the last digit was confirmed (ET mode annunciator flashes again!), the countdown can be activated by pressing CTL. Once it reaches zero, the alarm will go off like just described.

:information_source: There is a [YouTube video from A2A](https://www.youtube.com/watch?v=TrFYbS9yzuc) showing how it is operated (please note our simulation is more accurate to the book, however).
:information_source: The PR also adds a insim tutorial that walks one trough the functions in 8 minutes.

----
Open (minor) questions before merging this:
- is the alarm sound loud enough or too quiet?
- Should we add a tooltip telling that you can shift-click the SEL/CTL buttons?
  Currently this is only in the Documentation.md and tutorial description.
- [x] Some testing may be good, however I did already torture it (and will continue to do so the next days).

https://user-images.githubusercontent.com/13608602/113102698-a9b71200-91fe-11eb-9bdc-325c2b225b8b.mp4


hbeni:
I also added a 8 minute tutorial for fun. Currently Q/A, will upload shortly.

HHS81:
An aircraft model will be complex, if even the clock needs an extra 8minute tutorial! 😄 I will check

hbeni:
Haha, yes!
But it’s very easy once you know what buttons to press.

hbeni:
Just found a (very minor) bug:
- [ ] in ET countdown mode, when the alarm was confirmed using SEL, the countdown won't stop.

hbeni:
Fixed also the last todo (the annunciator shape is triangle instead of simple line).

So, if nothing comes up in further testing from you (which i doubt :grin:) , this is ready.

PR #410, opened 2021-04-12
Resized ASI.png (fix #409)

### LINKED ISSUE #409: Log error re: ASI.png
``` 
   6.12 [ALRT]:io         mipmapping: texture size not a power-of-two: /home/beni/Dokumente/FGFS/Aircraft/c182s/Models/Instruments/ASI.png
    6.31 [INFO]:OSG        
    6.31 [WARN]:OSG        PNG lib warning : Interlace handling should be turned on when using png_read_image
    6.42 [WARN]:OSG        PNG lib warning : Interlace handling should be turned on when using png_read_image
    6.56 [WARN]:OSG        PNG lib warning : Interlace handling should be turned on when using png_read_image
    6.61 [WARN]:OSG        PNG lib warning : Interlace handling should be turned on when using png_read_image
    6.86 [WARN]:OSG        Warning: Material::setShininess() of 254 is above permitted maximum, clamping to 128
```

It's exported as 1227x615 like the old one.

hbeni:
I resized the image to 1024x512 (from 1227x615) and the error is away.
Optically it's nearly unchanged.

@gilbertohasnofb Is this really neccessary? There where no visual errors or other problems before this fix, and I tought that the loader will resize dynamically; and besides that the texture should be only loaded once and then stay in memory?
So, is this really an issue we should merge, or should we live with the error in favor for a little better resolution of the instrument?


hbeni:
Mailing list (James) said, we should fix it:
> I'd recommend to fix it, unless you strongly cannot. Textures should always be power-of-two size, this is the easiest for the rendering side to work with. Non-power-of-two textures need some extra work, or waste some memory, depending on what happens internally.

I don't see such strong reasons, so I will merge this small fix now.

Issue #418, opened 2021-06-16
Multiplayer: Lights only correctly show on/off when local planes lights/power is enabled


hbeni:
Remote C182 light is only showed when the local electircal bus has power (and also the lights are switched on; but this was not tested dedicately; so I currently assume it's a power-only thing).

Probably the electrical system property is not correctly relayed and depends on a local value?

hbeni:
I could not reproduce this, when the local plane is _not_ an C182.
Lights work normal and as expected when viewed from the C172P and PA-28

hbeni:
For the icing, I maybe have found a problematic code piece:
https://github.com/HHS81/c182s/blob/ed9bb50ae9741d11b84c9a44aa98a84022804fc8/Nasal/c182s.nas#L512-L522

hbeni:
Hm, the lights problem did not appear in the [last session](https://forum.flightgear.org/viewtopic.php?p=392467#p392467):

https://i.ibb.co/ZB8F2sm/fgfs-20210922204625.jpg

TheFGFSEagle:
Yes - if there is a property named "dummy" in /ai/models/multiplayer[n], that will be used - if it's not there, FG will look for it in the global property tree of the remote instance.

hbeni:
So, this is a shader issue and cannot be fixed plane sided?

TheFGFSEagle:
Uh, it can be fixed plane side - just ensure that all properties the shader should use the remote value of are transmitted over MP ! :)

hbeni:
> Uh, it can be fixed plane side - just ensure that all properties the shader should use the remote value of are transmitted over MP ! :)

That is already the Case. All is working fine when you look at iced remote c182s when you are self _not_ in one.
As soon as you are in a 182 yourself, both your and the remote instances always shows the icing of your own plane.

TheFGFSEagle:
hum, that is weird … could you check on another aircraft maybe ?

hbeni:
What do you mean, what exactly do you echt to have tested?

TheFGFSEagle:
I want to know if you can observe the same behaviour replacing the 182 with another aircraft (172 or J-3 Cub for example) - if the issue happens there too, something must be wrong within core FG … but generally I don't know whatever could be going on here. It looks to me like the global propery takes precedence over the one in /ai/models/aircraft[n] which should not happen …

hbeni:
So we need to test this:
[...]

PR #424, opened 2021-07-02
Added MP slow properties

### LINKED ISSUE #420: Multiplayer: Wheels of remote planes are turning
The animation of the wheels are still tied to the local instance :)

### LINKED ISSUE #421: Multiplayer: Local referenced properties
OK. I've had the chance to look at the Effects, and I think the problem may be in Models/Effects/c182sreflect-uber.eff

Line 67:
      `<dirt-factor type="float" n="0"><use>/fdm/jsbsim/ice/graphic</use></dirt-factor>`

This is being set by `Nasal/c182s.nas` line 517.

My guess, as above, is that when you're using another plane, this isn't being set, and because it's undefined the value being passed into the Effect is wrong.   Certainly there's no way that should be an absolute path.

-Stuart

_Originally posted by @stuartbuchanan in https://github.com/HHS81/c182s/issues/417#issuecomment-863530909_

----
- [x]  fix local referenced property
- [x] make icing visible in MP (slow prop implementation)

### LINKED ISSUE #423: Multiplayer: PAX not showing in C182S remotes

hbeni:
Adding "slow MP properties" (TDM encoded string) for not so important props.

- [x] Fix #420
- [x] Fix #423
- [x] Fix #421
- [ ] <s>TODO #418 </s>

### DIFF
--- a/Models/Effects/c182sreflect-uber.eff
+++ b/Models/Effects/c182sreflect-uber.eff
@@ -64,7 +64,7 @@
 	<dirt-enabled type="int">1</dirt-enabled>
       <dirt-multi type="int">1</dirt-multi>
       <dirt-color type="vec3d" n="0">0.85 1.0 1.0</dirt-color>
-      <dirt-factor type="float" n="0"><use>/fdm/jsbsim/ice/graphic</use></dirt-factor>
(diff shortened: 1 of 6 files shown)


hbeni:
I now think we should merge this.
#418 should be tackled in a separate PR

hbeni:
I rebased this to current master to test some warnings i just got on the old branch

PR #432, opened 2021-09-06
Fix #431: Add keybind for rudder trim

### LINKED ISSUE #431: Add keybind for rudder trim
Right now the rudder trim could only be operated with the mouse and is rather coarse.
For fine adjustments it would be nice to have a keybind shortcut to operate the rudder trim.

I Propose to overload the default ones with <s>`shift+7`/`shift+1`</s> `crtl+4`/`ctrl+6`.

@wlbragg Do you know if there are such shortcuts for the C172 already? If so, i would want to use those for compatibility.

----
Has todo with https://sourceforge.net/p/flightgear/codetickets/2653/ (the keybinds in default will adjust `aileron-trim`, which we nullify locally here for the C182 variants, so only rudder-trim is applied. Currently the default keybind is broken however)

hbeni:
Adds keybinds `CTRL+4/6` to adjust rudder trim

### DIFF
--- a/c182s-set.xml
+++ b/c182s-set.xml
@@ -814,6 +814,14 @@
        <name>O</name>
        <desc>Open aerotow-hook</desc>
    </key>
+   <key>
+       <name>7/1</name>
+       <desc>Elevator trim</desc>
+   </key>
+    <key>
+       <name>CTRL+4/6</name>
+       <desc>Rudder trim</desc>
+    </key>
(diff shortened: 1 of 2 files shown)


hbeni:
Tested:
- :ok: with the patch to fgdata
- :ok: without the patch to fgdata

hbeni:
Should be fine to merge.

PR #434, opened 2021-09-11
Mariusz xc/springy strut

MariuszXC:
A comment in Systems/damage.xml: "Numbers MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for nose and main gear!" led me to investigate FDM setup. I discovered that damping coefficients were commented out, which may be a reason for plane 'jumpiness' on landings.

Also did a huge xml reformatting of FDM..

Testing was limited to one landing, there _seems_ to be an improvement.


HHS81:
@MariuszXC 
Thanks, but Sorry, but I can`t and I won`t acceppt this merge request.

1.) If the aircraft is jumping on landing you doing it wrong. The coefficients are correct, even agree with other commercial used sims.
And indeed the main gear strut on single engine cessnas are simple non-damped-springs- that´s why no damping. The comment you mentiond comes from a time in the past when the main gear struts were defined as damped spring - which is is actually wrong. 

2.) your merge-request contains changes which aren´t covered by your explanation. Why do we need a view-reset? Generally your merge request here seems somehow messy

Usually we open an issue first where name the problem, probably cause and discuss their fix- We want to simulate a high fidelity and realistic aircraft, and my experience in the last 6 years of developing this aircraft is, that many misconceptions about flying, technics, etc. are around.

hbeni:
> The comment you mentiond comes from a time in the past when the main gear struts were defined as damped spring - which is is actually wrong.

Hi Heiko, i will after this and remove the comment in master

hbeni:
> > The comment you mentiond comes from a time in the past when the main gear struts were defined as damped spring - which is is actually wrong.
> 
> Hi Heiko, i will after this and remove the comment in master

Done by ed9bb50

MariuszXC:
No problem and apologies for messy pull request. You were not supposed to even see  the view-reset, as it was supposed to be contained inside another branch I am experimenting with. Sorry for my low level of git-fu skills.

As for damping coefficient please note, that I did not touch the main gear which, as you wrote, is an undampened leaf spring. 
I was trying to address excessive rebounce of the front strut, which I believe is dampened. Maybe I am wrong though.

Greetings, Marius

hbeni:
Marius, is it OK if i cloise this PR and add an Issue, where we can discuss this further?

MariuszXC:
Yes, sure. This urned out not to be the correct approach to the problem anyway. 
(Formatting the XML gve me a chance to concentrate on the content)

PR #438, opened 2021-09-16
Fix #437

### LINKED ISSUE #437: DG: Tooltip should show current heading, not arbitary offset
Currently the "offset" knob of the directional gyro shows the offset setting of the knob.
This is not so helpful to synchronize against the compass.

Instead, the tooltip should display the heading the DG is currently pointing to.
This way, the pilot can more easily synchronize the DG to the magnetic compass.

----
- Ref #411

hbeni:
Screenshot_20210916_081253

Old (current) state:
grafik


hbeni:
Looking good!

PR #443, opened 2021-10-14
Implement Spark plug fouling

### LINKED ISSUE #439: Implement spark plug fouling
:information_source:  After reading up on this subject, I think it may still need some research.
Basicly there are two main sources of fouling, too rich mixture and too low combustion temperature:
- Carbon deposit fouling; the result of too rich mixture (incomplete combustion).
  They build up carbon at the spark plug which will eventually short the electrode.
- Lead buildup; the result of a too cold combustion which does not allow the [EDB lead scavenge agent](https://en.wikipedia.org/wiki/1%2C2-Dibromoethane) to do their job (temps < 264°F)

grafik

----

Running the engine in a "cold combustion state" (low RPM / rich mixture, available from JSBSim trough a AFR property) will result in carbon deposits on the spark plugs, causing a rough engine and loss of power, and can finally to a loss of the spark plugs ignition (magneto check can reveal this).
To remedy, the engine needs to run on higher RPM/hotter combustion for 15-25 seconds (POH 3-20 says "several minutes" for safety, but the shell letter and pilot forums also talk about the smaller timeframe).

To simulate this, we might do the following simplificated implementation:
(I *think* this can be preferably done in pure systems-xml-space (kinematic with variable speed setting?), however a nasal timer might be easier here. This should be further investigated, however)
- [x] Make the effects described here depending on the "spark plug icing" option enabled;
also extend the option label to include "fouling".
- [x] in the fuel system, calculate the present carbon deposit level, depending on the current AFR and engine temp
  - [x] Support two independently simulated magneto/plugs combos
  - [x] Determining by random number at starting time which spark plug(s) will be affected
  - [x] Make the other magnetos plugs slowly foul if running on just one mag
  - [x] <s>CHT as proxy? Or use</s> Revise the spark plug temp we introduced for the spark plug icing
  - [x] slowly build up in a mixture-rich and/or cold combustion regime (several minutes)
  - [x] quickly decrease in a hot/lean regime (20-60 seconds)
  - [x] (The system may probably be simplified to simulate only the final effect on the two magnetos; so summarizing the 6 cylinders into 1 for each magneto)
- [x] make flooded priming induce carbon deposits
- [x] Depending on the carbon deposit level, for each magneto calculate the effects
    - calculate a normalized power output ratio (0.0->1.0)
- [x] Calculate and apply RPM (and power) loss
  - _Goals_:
    - The goal here is to have a slight negative effect on power in cruise with spark plugs of a bad magneto.
   It should be generally better to fly with the *both* setting unless, as the POH says, the engine really is running rough (probably meaning alot of breaking effect of the bad spark plugs).
   - [x] If switching solely to the bad magneto, power should decrease noticeably.
   - [x] If switching solely to a good magneto, power should be slightly worse than on both (this is already simulated by JSBsim `<sparkfaildrop>` engine.xml definition)
    - [x] The second goal is to simulate the RPM drop effect at the magneto check, which should exceed the normal drop value.
      - _Note:_ The current lycoming-engine-xml says `0.15` power loss for a fully disabled magneto in "OK" status for the working one.
  - _Implementation idea_:
    - [x] It might be enough to calculate the loss-of-power based on the deposit state for each magneto (`0.15` foreach max).
    - [x] Based on this we can calculate for each affected magneto a "internal friction" value
     (this is somewhat hacky, but jsbsim does not have better hooks for that right now)
    - [x] Based on the magneto switch setting we can calculate a final "effect" value:
      - [x] if set to *left* or *right*: add the "internal friction" value of the respective selected magneto.
      This applies both the "fly on one mag" effect from JSBSim, as well as the "fly on fouled spark plugs magneto" effect, which is usually even worse.
      - [x] if set to *both*: we assume the better magneto has most of the load.
      -> calculate a weighted intermediary sum of the "internal friction" value of both mags: `(<betterMag>*0.85) + (<worseMag>*0.15)`
- [x] extend the "repair" function to remove deposits on the plugs
- [x] make the spark plug state persistent across sessions. (?Probably the effect is not strong enough to occur in a single session, and also descending seems to be a crucial factor (low-power descends), which can be seen in some aviation forums and is a factor for the next takeoffs mag check)
    
- [x] <s>I think that there is already a "rough engine running" sound. This may be adjusted and overlayed to the engine sound, getting stronger gradually with too rich/lean AFR and also a worsening spark plug fouling state (final friction value above?).</s> Nice idea for the future. Currently no sound. We do it via the coughing implementation for now.
- [x] If the "internal friction value" exceeds some threshold, rough running could also be achieved by utilizing the "coughing" effect we already have. This will worse the power too when running on one bad magneto.
- [x] Coughing effect may also be applied when the AFR is too lean/rich and combustion is nearly out of range
(subject to "complex engine procedures" enabled)
- [x] Reevaluate the "cold engine develops bad power" simulation: Probably this implementation here will be able to replace it => RESULT: It's still needed to simulate cold oil effects

- [x] Add documentation
  - [x] Documentation.md
  - [x] Tutorial showing the effect and how to remedy

- <s>[x] Add option for using unleaded fuel: https://www.greencarcongress.com/2021/07/20210728-g100ul.html
  - [x] extend fuel management dialog
  - [x] disable lead fouling when operating on AVGAS 100UL</s> postponed

- [x] Magneto malfunction": Make engine run rough when one magento failed
     (Optional, separate issue, not Spark Plug related, but easy to implement piggy-back wise)

----
References:
- Some more details in the c172p project: c172p-team/c172p#800
- https://pilotworkshop.com/tips/clearing_fouled_plug/
- Lyoming letter: https://www.lycoming.com/sites/default/files/Spark%20Plug%20Fouling.pdf
- On lead fouling: https://www.shell.com/business-customers/aviation/aeroshell/knowledge-centre/technical-talk/techart-18-30071600.html
- Again on lead fouling: https://tempestplus.com/wp-content/uploads/2019/01/Lead-Fouling-Written-By-AeroShell-Aviation.pdf
- Summarization by Dan Coles: http://flymafc.com/Articles/spark_plugs_fouling.html
- Another article, with video: https://www.cornerstonefbo.com/interesting-fact-about-aviation/often-asked-what-is-spark-plug-fouling-aviation.html
- P3D video showing effects: https://m.youtube.com/watch?v=GWCH9B5Q0D0

hbeni:
**Adding Spark plug fouling**
- Detailed spark plug fouling engine simulation
- Insim Tutorial to show the effect and how to avoid/mitigate
- [Documentation.md](https://github.com/HHS81/c182s/blob/Issue-439_SparkPlugFouling/Documentation.md#complex-engine-procedures) updated
- Added roughness if one magneto has failed

A overly rich mixture or idling on the ground for too long with cold combustion will foul your spark plugs, making the engine run rough. At the engine run up, the magneto check will reveal fouled plugs. There is a procedure to clear fouled plugs.

----
**Internals:**
- The engine coughing functionality was refactored in a generic "rough engine" implementation. Now the nasal part is very generic and mostly modelled in engine.xml and propulsion.xml. In the future, new "rough engine" sources can easily be added.
- There can be various sources for a rough engine, and a rough engine leads to oscillating RPM and loss of power.
- The spark plug fouling implementation is nearly purely done in propulsion and engine jsbsim systems and modular:
  The calculation of the underlying physics (spark plug temperature, fouling state, etc) is modelled separately
from the effect (->rough engine).
- The fouling state is saved with the plane's savestate and restored on the next session.
- The simulation was bound to the already available "spark plug icing" aircraft option which got renamed to ".../fouling". This is disabled by default if not explicitely activated from the pilot in the past.

- To express the effects, a new "power loss" effect is introduced; which is implemented currently by adjusting
a new friction source. If better options get available, we can switch the effects expression method easily.

- The systems are now also running on less-than-fcs fps to conserve CPU.

------
Fix #439

### DIFF
--- a/Documentation.md
+++ b/Documentation.md
@@ -44,8 +44,17 @@ Otherwise you need to pay attention to varios engine related things:
   the oil pump will not be able to supply enough pressure for good lubrication, leading to engine failure. Also monitor the oil service time as
   old oil does not perform so well (cycle it about every 50 hrs).
   Also some effects like additional friction with too hot oil and cold-start will be simulated.
-- `Allow Spark plug icing`  
-  If active, the spark plugs can freeze over, preventing engine start. This can occur in cold weather if the engine starts and quits shortly after, because then the engine is still cold, letting the water produced by the combustion process condensate on the plugs where they freeze over. This will prevent furter engine starts. To prevent this, preheat the engine properly, and if it happened, heat the engine so the ice melts (which can take considerable time!)
+- `Allow Spark plug icing/fouling`  
+  If active, the spark plugs can freeze over, preventing engine start. This can occur in cold weather if the engine starts and quits shortly after,
+  because then the engine is still cold, letting the water produced by the combustion process condensate on the plugs where they freeze over. This
+  will prevent furter engine starts. To prevent this, preheat the engine properly, and if it happened, heat the engine so the ice melts (which can
+  take considerable time!)  
+  Also, if the combustion temperature is too low and/or or the mixture too rich, the spark plugs can slowly foul with lead and carbon deposits,
+  eventually shortening the electrical spark. Be sure to keep the engine (CHT) warm enough by proper leaning and run-up procedures. In short, keep
+  the combustion temperature above about 264°C/507°F and avoid idling too long at <1000 RPM. Also, lean to peak EGT on ground/taxi and warm-up the
+  engine at 1200 RPM. Should a magneto check reveal spark plug problems, its most of the time some fouled plugs. Put the engine to 1800 RPM and 50°
+  lean of peak EGT. Let it run for about 30 seconds, monitor CHT. This should clean the deposits which a second magneto check should verify.  
+When parking the plane after flight, you should go to 1800 RPM and lean mix for 20 secs, then reduce throttle to 1000 RPM and then stop the engine by pulling mixture to the cut-off position.
 - `Winter Kit`  
   The winter kit is needed in cold weather (<20°F/-6°C) to reduce cooling air flow. If not supplied, the engine will possibly not get warm enough to develop good power, but be sure to remove it in hot weather, otherwise you risk too high CHT temps.
 
@@ -117,6 +126,8 @@ The pitot tube and stall horn can ice too, so you might not get any warning of a
 Some properties can be set at runtime to adjust the simulators state. You may use this for example in a script or trough a telnet session.  
 Additionally to the default ones, the c182 knows the following:
 - `/fdm/jsbsim/systems/propulsion/manual-friction` (in horsepowers) to induce additional engine friction, which reduces power.
+- `/engines/engine/manual-roughness-factor` (`0.0` to `1.0`) to induce engine roughness
+- `/engines/engine/manual-power-reduction-pct` (`0.0` to `1.0`) to make the engine less efficient
 - `/engines/engine/kill-engine` immediately kills the engine if set to `1`.
 
 ### KAP 140 Autopilot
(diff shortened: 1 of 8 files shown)


hbeni:
The new BSFC implementation for the engine roughness seems to do a better/cleaner job.

hbeni:
Extensively tested and I did not found something.

PR #445, opened 2021-10-15
Revise engine cylinder head mass

### LINKED ISSUE #444: Revise engine cylinder head mass
Currently the cylinder head mass is arbitarily set to 4 KGs. This influences engine cooling alot.

It was hard to find some source for the cylinder head mass, but I fund a replacement kit here: https://skygeek.com/textron-lycoming-lw13870-cylinder-assembly.html
It says 21.7lbs (~9.85Kg):
grafik

The entire engine weights (according to the list in the POH) 400lbs.
21.7*6=130, leaving 270lbs for other stuff.
Sounds sensible.

----
Flight testing showed that we now can do a max-power takeoff without quickly overheating in summer. Takeoff from 200ft at 24°C OAT (fair weather setting) allowed me to climb @80 knots to pattern altitude; CHT was 440°F.
Climbing further brought me to 2500ft when hitting the red line.

hbeni:
Previously we used 4Kg as arbitarily choosen head mass.
I found a shop listing a cylinder head at 21.7 lbs (~9.85 Kg; complete engine is 400 lbs according to POH 6-22).

Fix #444

### DIFF
--- a/Engines/engIO540AB1A5.xml
+++ b/Engines/engIO540AB1A5.xml
@@ -21,7 +21,7 @@
     <stroke unit="IN">4.375</stroke>
     <static-friction>1.5</static-friction>
     <cylinders> 6 </cylinders>
-    <cylinder-head-mass unit="KG"> 4 </cylinder-head-mass>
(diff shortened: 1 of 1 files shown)


hbeni:
I also made an inquiry at lycoming, however I do not expect them to answer me :(

Lets wait a bit and if not, we will go with the `21.75 LBS` value.

hbeni:
Just for the archives: Still no answer.

PR #447, opened 2021-10-18
Fix #446: Relocated invisible wheel block offset chock objects

### LINKED ISSUE #446: Securing: Placing chocks moves aircraft
Especially placing the main wheel chocks will move the airplane a small ammount.
Together with tiedowns this can make the plane "unstable" so that it will start to oscillate and bounce ("weird aircraft dance") in some conditions.

hbeni:
The offset was placed a little inside the wheel, making the plane do a bounce when initializing on chocks


hbeni:
@MariuszXC I tried to solve the "weird aircraft dance" again; and found that the wheel chocks offset object where slightly misplaced.
Now they are not in the wheel anymore, but very briefly in front of it. This solves also the moving/rotation of the plane with the right main wheel chock.
That should solve the long standing issue with the dancing aircraft (at least I hope so).

Could you please verify, as you were able to reproduce this fairly consistently?

hbeni:
@Octal450
Hi Josh, I need your help/input for a issue relating to PID-controllers, this time the tiedown control (which is the same for the c172 codewise as far as I could see).

I suspect that the PID controllers are off here. I have no Idea what is wrong, but maybe you have time to check the PID's values? It looks (to me) that the problem triggers when there are some huge interpolated values somewhere (maybe a big error calculated because of low frame rate?) No idea. It is tied to the systems performance however, as it only triggers at 8x speedup here for me (never a problem in 1x simulation speed).

https://github.com/HHS81/c182s/blob/8a00e4994c5d2c34b87454135cc5ec8dcda1d2b1/Systems/c182s-ground-effects.xml#L182-L191
(and some more, three in total, one for each tiedown)

hbeni:
(edited: make sure tiedows are applied, either form the aircraft->securing menu or by clicking on the clickspots at the wings and the tail hook)

HHS81:
> 
> 
> I do not have issues with C172, however I can get it to oscillate, but the procedure is much more involved. Same testing 
> location and command line, except for the aircraft, of course. When at runway:

I would need your weather settings as well, especially the turbulence settings in advanced weather. The wind in FGFS has some strange behavior on jsbsim aircraft

HHS81:
I use the same chocks on my 727-200adv- never had any issues with.
Well- you have to set the parking brake anyway, even with chocks you have to set them!

HHS81:
> 
>     * with [39a750d](https://github.com/HHS81/c182s/commit/39a750d56a4b467d342667cb563ef5d74d5c00dc): All is smoother for me. For you not.

I need to say: Only with tie-downs applied. Without tie-downs it is as you describe very well. That let me guess that there is something with the tie downs. Not the chocks are the issue- because even without it behaves strange. And my CPU is already very low. I used energy-save-settings and got only 14fps.

hbeni:
Yes, and especially the tail tiedown.
Lets wait if Josh finds something. Maybe it's just a bad offset from the nasal script placing the tiedowns.

HHS81:
@hbeni Wow! 😮 That works great- I don`t see any issues. And seems to me much more realistic! Great work, Benedikt!

Octal450:
@hbeni if you can solve this without PID, do it. PID is not the right tool for this approach. I would have done it your way also.

Kind Regards,
Josh

MariuszXC:
I tested the eef85f9 and try this:
(command line as in previous tests: ./run_fgfs.sh --enable-auto-coordination --airport=EPGD --timeofday=noon  --disable-real-weather-fetch --wind=280@4 --aircraft=c182s)
1. positioned on runway, in 'ready to fly' state
2. enable all three tiedowns (either via menu>securing or manually clicking hotspots)
3. give full throttle and.. enjoy the wild ride
4. pull throttle to idle, bouncing will continue 

It gets to the point when even menu>securing is unclickable, but once rear tiedown is removed bounce will slowly come to a stop.

PR #448, opened 2021-10-18
Add skydiving option

### LINKED ISSUE #440: Add Door removal option (Parachuters)
From the linked FAA article:
> The Cessna 182 is the most common jump plane in use

Add an option to remove the right door for parachuters to jump out :)

- [x] Add option in aircraft options dialog
- [x] <s>Adjust door action code to use `-1` as „removed position“</s> not neccessary
- [x] Hide model (door+window)
- [x] Hide backseats
- [x] Hide copilot yoke
- [x] Adjust multiplayer model to hide door/windows/yoke too
- [x] Add drag
- [x] Add forces to simulate jumpers on strut
- [x] Adjust air flow for cabin air cooling
- [x] Add drag sound increase
- [x] Adjust stances for PAX
- [x] Add trigger to let PAX jump out (and add submodels so there are actual jumpers jumping)
  Maybe keep this KISS and add clickspots to the PAX where when clicked, they jump?
  When jumping, set the PAX weight slot to `0` (this will also make the model disappear)
- [x] Port over to C182T as well

- - - -
References:
- https://www.faasafety.gov/files/gslac/library/documents/2016/Sep/120445/Flying%20for%20Skydive%20Operations.pdf
- https://diverdriver.com/cessna-182-skydiving-aircraft/

hbeni:
With the option enabled, seats, right door and yoke are removed.
When clicking on the passengers, they will jump out.
Alternatively all can be dropped at once using the "Jump!" button
from the aircraft options menu.

Fix #440

### DIFF
--- a/Documentation.md
+++ b/Documentation.md
@@ -145,8 +145,7 @@ The lower LCD row displays four different timing information, the active mode is
   - You can setup an countdown timer by by pressing the _select_ and _control_ buttons together (**shift-click one of the two buttons**). Enter the digits like just described at the FT alarm mode. After confirming the last digit, the countdown is armed and ready to be started by another press of _control_. The alarm will go off once the countdown reaches zero. You can confirm the the alarm by either pressing _select_ or _control_.
 
 
-Dual Control feature
--------------------------
+### Dual Control feature
 You can join a remote pilot as copilot! The pilot just uses the normal C182S Aircraft while the copilot must start the special "CoPilot" aircraft variant (select from the launcher).
 
 Once connected to the multiplayer server, both select each other trough the multiplayer menu entry *Select dual-control MP-pilot*:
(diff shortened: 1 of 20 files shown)


hbeni:
This is fun :)
Especially that the jumpers actually climb out to the strut (weight wise, no model doing this is shown so far) and this induces drag and weight.

Ready for merge I think.

hbeni:
I let it settle for a few more days, and then merge if nothing comes up.

> And also some fun!

I like how the plane drags to the right, when the jumpers climb out. I'm not aware that some other sim simulates this. This really is nice, however  really lack real world experience.
What i also miss is the special door, however i think we can leave that out since i read somewhere that you can get a STC that permits to fly without it, if everyone is properly belted.

hbeni:
Ah, okay, you already merged that, i missed this :)

Great!

PR #451, opened 2021-10-25
Parachuters are spawned in remote multiplayer instances

hbeni:
Remote instances now also drop the jumpers

(needs the detailed remote model loaded)


hbeni:
mpjumps_nasal

Converted the submodel implementation to nasal. This has the following advances:
- Local chuters are solid, too (for simulation of procedures and caretaking)
- Local instances land where remotes will see them too
- Support for freefall phase: Now the chute opens 5000ft AGL, freefall speed was matched to real world data (roughly, give or take a few seconds, but way close enough for our purpose) 

_____
ref: [1] https://ukskydiving.com/faq/what-height-does-the-parachute-open/

hbeni:
Local test worked fine.
Real MP testing still yet to conduct

hbeni:
(rebased to current master)

hbeni:
Merged - we tested this this wednesday evening with several planes.
That was fun!

PR #455, opened 2021-11-08
Adding Floats variant of c182s

### LINKED ISSUE #452: Add floats option
There are C182's outside in the world with floats, so we should have them too!
(Probably most of this can be borrowed from the c172 project)

- [ ] Add model + texture
- <s>[ ] Add option to switch to floats in aircraft menu</s> cancelled due to "separate variant" decision
- [ ] Add FDM ground reactions for floats
- [ ] Add drag effects
- [ ] Add water splash effects and sound
- [ ] Mooring option in aircraft securing dialog

------------
References:
- https://wiki.flightgear.org/Howto:Modelling_hydrodynamics_in_JSBSim

hbeni:
Mostly copied from C172 project

fix #452


wlbragg:
@hbeni
Is this anywhere near usable yet?

hbeni:
Basicly the ground effects should allow floating and landing on water, but this is far from complete

wlbragg:
I asked because I pulled it and the floats are defaulting to below the ground and the damage prop modeling is defaulting as visible and cannot be removed with the repair or disable of damage. I read your notes and expected the default landing gear to still be visible but not below the ground. Also the ground reaction on water is not responding. It basically sinks. Is there a certain way I should be putting this together or pulling the parts from the repository?

hbeni:
Well, this branch was a try to port over the implementatoin from the c172p, however I believe I made a mess.
Probably the easiest would be to again start fresh, port only the model, and reimplement the systems.

When I tested the latest code here, I was able to float on the water unharmed.

This effort is more that I can deliver ATM, however, so if we want floats, someone more capable than me needs to step in.

wlbragg:
Okay, no worries. I was just curious. If I get some time maybe I will have a go at it. It really shouldn't be too difficult, but I agree there are a lot of pieces. You've done so much to this aircraft I would love to see this variant as well. What about skis, have you seen any real life 182's with skis?

hbeni:
No skies seen so far, but otoh i didn’t search for that yet.
I just had seen one on floats (unclear if amphibious!) when i did research the new liveries a short while ago, and given we had already done alot of wednesday-ga flights with water landings I thought it would be nice to have the option.
I was inclined to try out if i could port it but failed

wlbragg:
Yeah, I did a quick look at what you did.
I know you want to keep this a true variant and not load it with the default version to keep the load down. But to do that you have to break out he default gear so each variant only loads the needed gear.
Variant default loads gear model.
Variant float loads pontoon model.
That, as you know, is not how I did it.
I don't think what you want is that difficult, it would just require we make the default gear a separate xml file and only load it when needed. Same as the floats.
Do I have this right?

hbeni:
> Do I have this right?

Probably :)

TheFGFSEagle:
Actually you don't need to make the default gear a separate file - you can just set a property, say /sim/model/floats-enabled, in the c182s-floats-set.xml file and then do a select animation on both the default gear and the floats with a condition on that condition !

wlbragg:
Yes, that is the way the c182p and the j3 do it. But the c182 developers don't want the added weight (as in overhead) of the float or pontoons always affecting the default configuration. They would rather it be a completely separate variant. Which is really not any harder to do, you only use the functionality of switching without restarting. But you gain in other ways. If there is a damage model then you have that overhead as well. If you decide to add skis, then even more. I've debated in my head many times which way I should go on the aircraft I developed. I always went with the select animation, but it would have been ok the other way also.

hbeni:
The question is, if we combine the floats, and also it’s systems logic into the base model, how much burden does it add to the fps?
Or is there a way, we can just perform the hydrodynamics calculations when actually on floats?
And do unselected 3D objects need calculation time?

TheFGFSEagle:
I don't know if hidden objects take calculation time, but I don't think so - there are no vertices drawn, effectively. 

As for the hydrodynamics, whole JSBsim system channels can be switched on or off based on a property. But I think it's better (and simpler) to keep it a true variant (e.g. not in-sim change of the floats / normal gear) - it's more realistic IMO.

I've cleaned up the mess in my local copy of the repo - can you give me write rights here, or would you prefer I create a fork and push to that ?

There are some things left to do, but I hid the main gear, placed the floats in the right place, adjusted the float bogeys and structure contacts so the floats aren't below ground, textured the float wheels, fixed the float wheel animations and coupled the float gear lever to the float wheels - it's flyable (and water land-able), but some things aren't right still - there are two pieces of the main gear strut on the fuselage which should be transferred to the main gear struts, and I couldn't get the water rudders to extend / retract. Also the propellers aren't spinning, but at least the correct one is getting selected. I'll look into that, but first I'd like you to take a look at my changes (things I missed, etc.).

HHS81:
@hbeni @wlbragg @TheFGFSEagle 
Hey guys, 

I was aware of the floats for the C182S since I started the aircraft, but it wasn`t my main interest, though I found it interesting. 
The reason to have a sperate model of the c182s with floats is first that it is recommended by the core developers for inclusion into FGAddon. Second, it will getting more and more complex to maintain the aircraft, with all of its systems, additional models, nasal scripts, ... Another reason- for the float variant you will need more powerful engines.

Hidden objects doesn`t take calculation time. But whenever a object is loaded it will need, and especially with textures applied, and scripted by nasal script. I notice it everytime I crash the c172p- when the wing, gears etc. changes to "crashed", fgfs freezes for a tiny, but noticeable moment. And the c172p has very, very poor framerates compared to the c182s. When I started the c182s my goal was to have a very realistic aircraft with great fps without loosing realistic graphics and systems. The way the c172p does it, is not the way I want to have it on the c182s. It will kill perfomance, and it is not like to be wanted in FGAddon as I understood some time ago.

Btw. The floats on the C182s are retrofitted, Wipaire is one of the companies selling them: https://www.wipaire.com/aircraft_page/cessna-182-skylane/

HHS81:
The drawside of the type of configuration is, that it will increase package size a lot.

HHS81:
>>I don't think what you want is that difficult, it would just require we make the default gear a separate xml file and only load it when needed. Same as the floats. Do I have this right?

Why so complicated? 

Simply have:
c182s-floats.xml
c182s-floats-set.xml
Models/c182s-floats.xml with a simple select animation of the main landing gear objects and a select animation of the floats.xml which contains the 3dmodel of the floats, gear levers etc.  This doesn´t break anything, doesn`t need much work and friendly to computers 😄

TheFGFSEagle:
Well, I think I combined both - there is just a slight increase in package size because there is no completely separate model for the floats version, yet the floats are not loaded at all with the default variant, only with the floats variant. I'll create a fork and push my changes there so that you can take a look and tell me what you think about it.

hbeni:
Good idea with the fork.
Can you also make it easy to add skis? And bush wheels, while we are at it? :)

TheFGFSEagle:
As there are a few things not working on the floats variant (and I got to finish a scenery in the next days !), I'd like to do the skis much later. The bush wheels shouldn't be that hard, so these I could do soon.

hbeni:
No rush! This is all not as easy, as it first looks - I already acknowledge that from first hand experience :)

The thing I want to highlight is that we really should find a way to not add framedrop and also to come up with a extensible solution (so floats and skis are easy to add).

TheFGFSEagle:
I did not notice any frame rate drop between the default variant and the float variant. And the way you began it and I now continued it it should be fairly easy to add the bush wheels and skis too. Just take a look: github.com/TheFGFSEagle/c182s, the Issue-452_Floats branch.

wlbragg:
Note: I don't know if things have changed since @HHS81 tested the difference in framerate between the 182 and the 172. But I remembered testing myself and that there was no distinguishable frame rate difference between the two aircraft. I tested again and in the default gear configuration they are identical. The amphibious version of the 172 loses about 2 fps. I'm not posting this to change any opinions on how to proceed with this, but merely as a data point for consideration. The discrepancy between HHS81's observation and current may have been when the 172 also had the fg1000 included and not as a separate variant. That was a huge resource hog and seriously affected frame rate. 

>Why so complicated?

Also to clear this up. Either I didn't explain myself correctly or there is some confusion here. I would likely use a separate xml file for the two different configurations (amphibian and default). Still using a select animation as to which file to load or include it in the appropriate set variation file. The reason being is the amphibious configuration requires independent gear animation, float rudder and gear control logic and it need to be included somewhere. I think it is cleaner to contain all that logic in a separate xml that can be loaded. It is organised and isolated that way. I don't think this wasn't any more or less complicated. If I were to do this from scratch I would do all modeling in their own independent blends. The blend files can get rather complicated otherwise and someone coming in behind the original author would likely find it easier to comprehend what is going on.

I learned a lot after creating all the variant 172 layouts. For the most part I think in the end I got it right. I refactored things more than once. But for sure there are things I might have done slightly different if I were to do it again. As far as complicated goes, it is complicated no matter how it is done in the respect that there are many areas that need to be touched to add a variant. Potentially including, electrical, systems, replay, weighs, drag, bogie positions, modeling interior and exterior, hydro code and possibly damage model and system if so inclined.

Anyway, I'm speaking about this from the benefit of having the experience doing this a couple times on a couple different aircraft.

TheFGFSEagle:
I've no idea anymore how it is currently on this branch - but yeah, floats / normal gear each in their own files definitely makes sense.

I cannot believe my last message in this issue is from over 2  years ago … how fast can time pass ?

hbeni:
> I've no idea anymore how it is currently on this branch

No changes since....

> 2 years ago … how fast can time pass ?

Floats would definitely be very cool.

PR #458, opened 2021-11-24
New Davtron803 texture

hbeni:
Old:
grafik

New:
grafik
grafik

Davtron Hires photo texture taken from the DA40NG project


hbeni:
Made the LCD background a little more green and removed the reflections:

grafik

grafik

grafik

hbeni:
Today I reverted that change. There seem to be different versions of the clock face, and the one we had before the change is (according to some pics I found) the one installed in the 1996 C182s.

PR #461, opened 2021-12-01
Better oil pressure simulation

hbeni:
Cold oil will now more slowly build pressure.

- oil pressure depending on viscosity
- viscosity depending on oil grade and oil temp

:point_up: Potentially can support switching oil grades in the future! :)

---
**Sources**
- POH 4-21:
  grafik

- POH 8-14:
  grafik

- 15W-50 viscosity specs: https://www.mobil.com/en-us/passenger-v ... il-1-15w50

- 15W-40 viscosity courve: https://wiki.anton-paar.com/us-en/engine-oil/

- some real life reports suggesting the current simulation is somewhat realistic (takes some time (minutes in very cold weather) where the needle moves slowly): http://www.cessna172club.com/forum/ubbthreads.php?ubb=showflat&Number=102858

### DIFF
--- a/Engines/engIO540AB1A5.xml
+++ b/Engines/engIO540AB1A5.xml
@@ -36,5 +36,5 @@
     <oil-pressure-relief-valve-psi> 67 </oil-pressure-relief-valve-psi> <!-- lycoming manual states: min=55; max=95; idle=25; start&warm-up=115 -->
     <oil-pressure-rpm-max>          1000  </oil-pressure-rpm-max>
     <design-oil-temp-degK>           358  </design-oil-temp-degK>     <!-- level cruise: between 165degF and 200 degF -->
-    <!--<oil-viscosity-index>              0.25 </oil-viscosity-index> --> <!-- TODO: .25 default jsbsim value -->
+    <!--<oil-viscosity-index>              0.25 </oil-viscosity-index> --> <!-- TODO: .25 default jsbsim value, but simulated in engine.xml -->
 </piston_engine>
--- a/Nasal/c182s-states.nas
+++ b/Nasal/c182s-states.nas
@@ -106,8 +106,6 @@ var setEngineRunning = func(rpm, throttle, mix, prop) {
     #
     # Engine start function
     #
-    setprop("/sim/start-state-internal/oil-temp-override", 1); # override disables coughing due to low oil temp
-    settimer(func{ setprop("/sim/start-state-internal/oil-temp-override", 0); }, 240); # disable override after this time
     
     # Ensure disabled complex-engine-procedures
     # (so engine always starts)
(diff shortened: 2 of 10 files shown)


wlbragg:
Out of curiosity, this is only a gauge indicator that this is affecting and not performance? What kind of time are we looking at in really cold weather? In cold weather could it be long enough for a dumb pilot to try to take off with bad oil pressure? Can that be incorporated into the performance of the aircraft? What about into damage to the engine over time, like low oil, putting the aircraft under stress by throttling up before the oil is warmed? I am not suggesting you necessarily go that far, just wondering if any of this is something you considered.
Even if not, it's still a nice visual and adds to the reality of waiting for the OP to come up in really cold weather. I like it!

hbeni:
> Out of curiosity, this is only a gauge indicator that this is affecting and not performance?

Yes, that part is just visual for now. More on that below…

> What kind of time are we looking at in really cold weather? In cold weather could it be long enough for a dumb pilot to try to take off with bad oil pressure?

Yes, very well. I turned it to the POH, so we talk roughly about 15 seconds at summer time and approaching a minute at -6 degC.

> Can that be incorporated into the performance of the aircraft?

This was already modeled based on other factors, mostly oil temp. So if you have good pressure but the oil is still cold, you have not all available power, and in worse situations you may even kill the engine. It also may run rough.

> What about into damage to the engine over time, like low oil, putting the aircraft under stress by throttling up before the oil is warmed? 

I have tought of this already alot, and it would indeed be a nice feature. What basicly already happens is that we simulate additional internal engine friction, and that could probably easily be leveraged into such a feature.

> Even if not, it's still a nice visual and adds to the reality of waiting for the OP to come up in really cold weather. I like it!

Thanks!

hbeni:
That said i need to crosscheck with the starting ready for takeoff state

Octal450:
@hbeni , I strongly suggest porting these things into a JSBsim system.

Kind Regards,
Josh

hbeni:
> That said i need to crosscheck with the starting ready for takeoff state

So, after some more work than anticipated, I was able to get rid of an old workaround and refactor the engine temperature stuff a little.
The "start with engine on" states now do add initial oil and cht temps, so the too-cool-oil effects are not triggering.

Should we ever get JSBSim to initialize at OAT changes and compensate for that itself at startup, we can easily remove the "workaround" by just removing the compensator nasal+jsbsim systems code, which is (hopefully) trivial to find now (in fact, the compensation could just be removed by adjusting the new `<summer>` which calculates the final temperatures).


I already had some extensive tests, but a little more testing should be done; especially when starting in cruise and cruising for about two minutes (two minutes is the time the initial temp-add-offset is adjusted)

hbeni:
Some more testing yields good values. I think this is OK for merge.

PR #468, opened 2021-12-17
Add AI wake turbolence

### LINKED ISSUE #462: Implement wake turbolence from AI/MP planes
https://wiki.flightgear.org/AI_wake_turbulence

Seems to only be possible with AI aircraft.

Maybe we could add another wake external reactions set and calculate that ourselves. However - this should in my opinion be something generic, so is really a FGData (or even core) thing to implement. It should not be C182 specific.
There is/was a hackatron effort: https://wiki.flightgear.org/Hackathon_Proposal:_Wake_from_multiplayer_aircraft

hbeni:
Adding the jsbsim hooks to support AI wake turbolence.
MP wake turbolence is currently not supported by fgfs, but probably can be expected to use the same hooks in the future.

(Fix #462)

### DIFF
--- a/c182s.xml
+++ b/c182s.xml
@@ -2144,6 +2144,36 @@
             </direction>
         </force>
         
+        <!-- wake turbulence from other AI plane (direction is controlled by properties automatically) -->
+        <force name="ai-wake" frame="BODY">
+            <location unit="IN">
+                <!-- Center of the main wing leading edge -->
+                <x>32.0</x>
+                <y> 0.0</y>
+                <z>52.5</z>
+            </location>
+            <direction>
+                <!-- These are dummy values. They are ignored by FlightGear.
+                     Their purpose is to avoid JSBSim display warning messages
+                     about <direction> not being specified.
+                     -->
+                <x>1.0</x>
+                <y>0.0</y>
+                <z>0.0</z>
+            </direction>
+        </force>
+        <moment name="ai-wake" frame="BODY">
+            <direction>
+                <!-- These are dummy values. They are ignored by FlightGear.
+                     Their purpose is to avoid JSBSim display warning messages
+                     about <direction> not being specified.
+                     -->
+                <x>1.0</x>
+                <y>0.0</y>
+                <z>0.0</z>
+            </direction>
+        </moment>
+        
     </external_reactions>
     
     <system file="c182s-stallhorn"/>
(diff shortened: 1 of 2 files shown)


hbeni:
Happened also on the C172P in munich (where i have Ortophotos deployed).
Tested also at Keflavik with default scenery. Breaks too.
Really looks like a core thing, so:
- opened ticket https://sourceforge.net/p/flightgear/codetickets/2681/
- merging this

wlbragg:
@hbeni 

I was looking at the c172p to make sure we also had implemented this. But the comment in the code...
>Center of the main wing leading edge

Are not what the c182 or the c172 are using.

We are using this...
Capture

When according to the comment in the documentation, we should be using this...
Capture2

Did you check out any documentation that led you to believe we have it correct?

hbeni:
Uh, no, i trusted your numbers :)

wlbragg:
OK, I don't even remember doing this and after searching for it...

https://github.com/c172p-team/c172p/pull/989

wlbragg:
The existing numbers may be right, I still don't understand the difference from the model position and the FDM frame of reference. I guess I don't understand why anyone wouldn't make them the same when setting up the model in Blender. Otherwise, you just making everything harder for yourself and everyone that follows. It is not something I would do I it were up to me. If they are so different, then most anything I ever measured for weight point or other features that influence FDM are probably wrong!

wlbragg:
In what world does that first image's cursor position make sense for the FDM's center point for the leading edge of the wing?

wlbragg:
@hbeni from what I can tell the 182 is also crashing immediately as soon as the AI wake effect is influencing the FDM. So as far as I am concerned, it is completely broken.

hbeni:
> @hbeni from what I can tell the 182 is also crashing immediately as soon as the AI wake effect is influencing the FDM. So as far as I am concerned, it is completely broken.

Yes, but I think it is caused by the calculations in fgcore, not the model.
It may be that or coordinates make that haben, but in my opinion that should never be allowed to cause a crash.

wlbragg:
Yes, it's the core code. But from what I could tell it is completely broken, not only sometimes in all implementations I tried so far. I also tested the Saab 37. I've yet to actually see the effect in action because it crashes the second it is activated. I posted this finding to the ticket you made.

PR #472, opened 2021-12-20
Implement Flap damage due to overspeed

### LINKED ISSUE #463: Extending wing flaps at too high airspeed should get them stuck / rip them off
May be simulated easily with:
- [x] getting stuck -> triggering `serviceable`
- [x] ripping off -> 
  - [x] hide model (also for MP)
  - [x] add negative lift coefficient to simulate reduced lift
- [x] Maybe we then can get rid of the message from the simple limits framework
- [x] Make repairable

POH 2-4/2-5:
grafik
grafik
POH 7-13 describes the wing flap system in detail.
grafik

----
- May be also done together with #464
- some reports on consequences: https://www.pprune.org/pacific-general-aviation-questions/260242-flaps-question-you-cessna-fanatics.html

hbeni:
fix #463


hbeni:
Should be good, but I think some more testing would be nice

hbeni:
Tested by Ysop, confirmed by me

PR #475, opened 2022-01-13
Implement 474 (in air starts need a stabilizer)

hbeni:
This is implemented by a simple "autopilot", that is only active the first few seconds of the sim when starting in air.
It controls the yoke and tries to counter the forces of the engine startup (and also wind forces etc).
It can be disabled by supplying `--prop:/sim/inair-startup-stabilizer=0` in the launcher.

Together with this the "approach" state was adjusted slightly, so now prop and mix levers
are in the (final) approach setting from the checklist. Throttle was reduced to 50%, and
landing lights are now also switchedswitched on.


hbeni:
Ready for testing :)

hbeni:
@HHS81 What do you think of this feature?

PR #477, opened 2022-01-19
Add pedestal light as compositor light spot.

### LINKED ISSUE #476: Trim wheels not lit
Currently the trim wheels are not lit by any light.
They only are affected by ambient lighting. That makes them hard to find at night.

grafik

_____
Ref:
- https://github.com/HHS81/c182s/pull/260 This draft-PR already discussed this issue but found no good solution
- https://github.com/HHS81/c182s/issues/176 Brought up the idea to use a compositor light

hbeni:
This way the trim wheel is lit and can be seen in the dark
The light is controlled mainly by the pedestal light knob but also affected by the glareshield and dome lights.

Fix #476

### DIFF
--- a/Models/c182s.xml
+++ b/Models/c182s.xml
@@ -33,6 +33,9 @@
         # Note: This nasal code is loaded just for remote multiplayer models
 
         var rplayer = cmdarg();
+        
+        # mark model as remote
+        rplayer.getNode("sim/model/c182s/remote-instance", 1).setBoolValue(1);
 
         # Set up property aliases for animations.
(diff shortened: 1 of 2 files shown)


hbeni:
@HHS81 What do you think about this solution?

grafik

wlbragg:
Did you check to see that it doesn't bleed through to the outside of the cockpit? I tried using it in the AirCrane, but the size and intensity of the light I needed bled through the skin of the cockpit. That example looks small enough it might work.

hbeni:
Yes, i restricted the light range to very few centimeters. For me it does not bleed trough.

PR #478, opened 2022-01-20
Fixes/tewaks for radio stack / instrument lighting

hbeni:
At nighttime, the radio stack and instrument lighting should briefly illuminate the cabin.
The current effect was so faint, that it was not visible in most cases.
Therefore this commit makes the effect a little more prominent.

Also, the radio stack lighting was not correct:
- the COM radios were not adressed correctly and thus never added light to the mix
- Some equipment is not lit when the radio stack light knob is off
- COM units should have a bigger effect on brightness than other equipment (it has more red lights)

### DIFF
--- a/Systems/als-lights.xml
+++ b/Systems/als-lights.xml
@@ -221,7 +221,7 @@
                 <sum>
                     <property>/sim/model/c182s/lighting/audio-panel</property>
                     <property>/sim/model/c182s/lighting/dme</property>
-                    <property>/sim/model/c182s/lighting/nav-1</property>
+                    <property>/sim/model/c182s/lighting/nav-1</property> <property>/sim/model/c182s/lighting/nav-1</property>  <!-- counted twice because of com+nav unit -->
                 </sum>
             </expression>
         </input>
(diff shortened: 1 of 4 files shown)


hbeni:
Hm, not sure, maybe the effect is too much now. However, I would assume that the dark-adapted eye would sense the backlight scatter rather intensily.
But OTOH the lit gauges should make the pupil more narrow, effectively reducing the effect.

I'm undecided. the good thing is it's easily tunable.

hbeni:
Sky was covered so nearly no light (had to use the flashlight to find the battery knob)
Consider viewing in full-screen to avoid other bright surfaces.

Midnight, full Radio stack light intensity:
grafik


Midnight, full instrument light intensity:
grafik



Midnight, full Radio stack and instrument light intensity:
grafik

TheFGFSEagle:
Compared to the bright instrument lighting, I'd even say the radio stack lighting is too dark - but I've no experience with this.

hbeni:
You talk about the light reflection on the panel?

MariuszXC:
Looking at radio stack alone, the AP display is too dim compared to buttons and other displays around it (probably has to do with scotopic vision sensitivity).

Comparing radio stack as a whole with instrument panel as a whole, I'd say that instrument panel is too bright.

Would be ideal to have a night photo of a real cockpit, with brightness set to maximum and compare..

hbeni:
Tuned the light values down again.
To clarify: The question/tuning here is adjusting the panels scatter light, ie the light ammount that the radio stack and instrument lights reflect back into the cockpit. It's *not* regarding the gauges/radio stack light itself, but the reflection effect in the cockpit/panel.

Issue #67, opened 2016-02-02
Idea: spiral cable from panel to yoke 


HHS81:
adding a spiral cable from panel to yoke,  animated by scale-animation; using control-texture

Background:
Cable in Real Life goes to the map light mounted at the bottom of the yoke. Not possible in default or ALS, but able in Rembrandt.

TheFGFSEagle:
It's one of the next things I was going to make for my Cessna 210 - so I could probably copy that to the C182S with a few modifications - I guess I should create a fork of this repo and make a PR when I'm done ? If there's still interest in this. ;)

hbeni:
Yes, there is interest; and yes, clone the repo and do a pull request :)

TheFGFSEagle:
Okay ! In fact, I have a fork already … 😆

PR #486, opened 2022-04-19
Implement #67 - spiral cable from yoke to panel

TheFGFSEagle:
Added spiral yoke map light cable, and (hopefully) fixed yokes graphics glitch for some graphics cards (https://forum.flightgear.org/viewtopic.php?f=4&t=28475&start=15#p395288) by assigning `model-transparent-combined` to the yokes + yoke cables


HHS81:
> Added spiral yoke map light cable, and (hopefully) fixed yokes graphics glitch for some graphics cards (https://forum.flightgear.org/viewtopic.php?f=4&t=28475&start=15#p395288) by assigning `model-transparent-combined` to the yokes + yoke cables

From the effect file: "It's kept for backwards compatibility and should not be used on new projects." Can you tell me why you did assigne "model-transparent-combined" to the  yoke?
Your explanation link is leading to an old forum thread with nothing about the problem...

TheFGFSEagle:
Well, making the cable come out of the panel directly below the yoke mount should be fairly easy ! The bend, though, I've tried already by cutting the cable in three pieces (or actually, 5, but the first and last just move with the yoke / are attached to the panel) but I had some trouble making the animation of the cable fit the animation of the yoke (on my 210). But I can retry by using just two pieces with a locked track animation … that should work. But the cable would get an really sharp pointing tip in the middle ?

hbeni:
Hm, as close to the footage as possible would be nice.
Didn’t the c172 have a bent already?
For sure also the c208 grand caravan

TheFGFSEagle:
For the glossyness, I guess I can use the c182s_controls effect ?

TheFGFSEagle:
The 172 has a very small bend in the cable AC file, and that just get's scaled. As for the 208's yoke cable, it's also just one piece that gets scaled … :(

TheFGFSEagle:
I'll see if I can get something better, with three parts and two locked track animations. But actually, this is vertex shader territory … but unfortunately I don't know that AT ALL !

HHS81:
> I'll see if I can get something better, with three parts and two locked track animations. But actually, this is vertex shader territory … but unfortunately I don't know that AT ALL !

Have a look into wlbraggs Aircrane- he animated the hose. The animation is quite similar to that needed for the cable. Nom need for vertex shaders

hbeni:
Awesome bending!
Can you tune the bending a little more, so its bend somewhat when the wind gust lock is installed, like in the picture?

TheFGFSEagle:
You mean, that it has a little bend already when the yoke is in the neutral position ?

TheFGFSEagle:
@HHS81 sorry about not replying to your comment - I completely overlooked it !

> > Added spiral yoke map light cable, and (hopefully) fixed yokes graphics glitch for some graphics cards (https://forum.flightgear.org/viewtopic.php?f=4&t=28475&start=15#p395288) by assigning `model-transparent-combined` to the yokes + yoke cables
> 
> From the effect file: "It's kept for backwards compatibility and should not be used on new projects." Can you tell me why you did assigne "model-transparent-combined" to the yoke? Your explanation link is leading to an old forum thread with nothing about the problem...

Well, the post itself doesn't contain anything about the yoke transparency problem, but the video included _does_ ! As to why I assigned the effect - to fix exactly that problem, as I was getting it too. I didn't know it was deprecated, but I removed it again anyways as it didn't change anything. May I ask why we don't simply hide the yokes, like the C172P ?

TheFGFSEagle:
There you go !

hbeni:
> May I ask why we don't simply hide the yokes, like the C172P ?

Because :)
You *can* hide them completely if you adjust the transparency setting from the crafts menu.
The option enhances user experience for people like me that still want feedback where the yoke is but also want to see the panel behind it.

hbeni:
> @hbeni: There you go !

Will investigate soon

hbeni:
Looks good, thank you very much for the addition! :)

HHS81:
> 
> Well, the post itself doesn't contain anything about the yoke transparency problem, but the video included _does_ ! As to why I assigned the effect - to fix exactly that problem, as I was getting it too. I didn't know it was deprecated, but I removed it again anyways as it didn't change anything. May I ask why we don't simply hide the yokes, like the C172P ?

The linked thread has already more than 13 pages and several videos. 
Since you are new to FGFS a little rule I had to learn myself years ago when I started with FGFS: if you think it is a bug- it is in fact a feature. 
😸

TheFGFSEagle:
> The linked thread has already more than 13 pages and several videos.
It's on the last / first page ! ;) Also the link is a direct link which should take you directly to the right post with the video …

HHS81:
> > The linked thread has already more than 13 pages and several videos.
> 
> It's on the last / first page ! ;) Also the link is a direct link which should take you directly to the right post with the video … (the description of the post only says something about a lighting problem, but you can also see the yoke problem in it)

No, it´s not. That`s the link you posted: https://forum.flightgear.org/viewtopic.php?f=4&t=28475&start=15#p395288
Links to the second page of the thread (from 29. dec. 2015)

And that´s the link you should have posted: https://forum.flightgear.org/viewtopic.php?f=4&t=28475&start=150#p395286 (from nov. 2021)

Please compare it. 

The yoke problem is a problem inside FlightGear, not the aircraft itself. There is the rule that in FlightGear we won`t make changes to aircraft just for fixing issues caused by internal bugs. 

The lighting problem of the C182T panel is a configuration problem of the effect file combined with the 3d-model.  The prerendered lightmaps are not matching the panel since it changed, and should be rerendered.

TheFGFSEagle:
@HHS81 okay, now I get it … I have reversed my sorting order so that I get the newest post on the first page, instead of the oldest … I didn't know this makes a difference when sharing a link to a reply. Sorry.

HHS81:
> @HHS81 okay, now I get it … I have reversed my sorting order so that I get the newest post on the first page, instead of the oldest … I didn't know this makes a difference when sharing a link to a reply. Sorry.

Nether mind. I´m pretty sure it should not make any difference. Maybe a bug in the software used by the forum?

PR #493, opened 2022-11-02
Issue 492: kap140 update

### LINKED ISSUE #492: Update KAP140 upstream
Sascha pushed an update to the KAP140 which brings better hardware integration:
See https://github.com/reiszner/KAP140/issues/13

We should integrate it as test case for testing.

-----

- [x] In the kap140.xml, at the end there is a reference to the PA28, breaking display of numbers
- [x] CWS is always shown as disengaged (the "AP" in the box should vanish if pressing CWS button)
- [x] In VS mode, I cannot change the climb/descend setting with single clicks (holding the buttons works tough!)

Tested positive:
- [x] ALT (with and without capture), VS mode
  - [x] ALT Mode altitude change by single UP/DN button presses
  - [x] ALT mode continuous alt change by pressing+holding UP/DN button
  - [x] VS mode incrementing speed setting by pressing+holding UP/DN buttons
- [x] HDG mode
- [x] ROL mode
- [x] ROL->NAV catch
- [x] ROL->ILS catch
- [x] HDG->NAV 45° intercept
- [x] REV mode, from ROL, HDG and NAV
- [x] ILS APR (including 180° outbound tracking - 16 knots wind from left side!)
  - [x] GS capture and trackig worked as expected
- [x] CWS actions in various configurations (including changing altitude in NAV/ALT mode combined with evasion, like ATC may request)
- [x] Overflying VOR

hbeni:
Refactored KAP140 to more easily interface with external hardware etc

fix #492

### DIFF
--- a/Models/Instruments/Avionics/kap140/kap140-properties.xml
+++ b/Models/Instruments/Avionics/kap140/kap140-properties.xml
@@ -21,6 +21,10 @@
             <hsi-installed type="bool">false</hsi-installed>
             <baro-tied type="bool">false</baro-tied>
         </config>
+        <event>
+            <button-cws type="bool">false</button-cws>
+            <button-disc type="bool">false</button-disc>
+        </event>
         <panel>
             <state type="int">0</state>
             <state-old type="int">0</state-old>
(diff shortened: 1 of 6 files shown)


hbeni:
Some final testing, then we can merge

PR #496, opened 2022-11-16
Add compass correction card (texture by @hbeni)

### LINKED ISSUE #495: Magnetic Compass: Add deviation table
https://en.wikipedia.org/wiki/Magnetic_deviation

The compass has magnetic deviation, and this must be compensated using a table.
That table is usually shown below the compass.

The table is here:
https://github.com/HHS81/c182s/blob/8befd6b94d2087340dbb522453c88474ae75c919/Systems/instrumentation.xml#L93-L150

----
The DA40NG has such a card

TheFGFSEagle:
fix #495


TheFGFSEagle:
Done ! (for the 182S - if you want me to add it for the T too I need a photo for the correct placement).

hbeni:
Hm, i think the placement is not standardized, so we could put it where we want.

Which brings me to an idea, can we easily move it more to the topleft and make it 75% the size (the original is even smaller)?
grafik

hbeni:
Checked the lights, looks good!
One small thing is when activating the cabin lights (the ones at the top) it does not illuminate the card

grafik

TheFGFSEagle:
Moving and resizing is no problem at all ! :) The thing with the lighting is more complicated however … does anyone still have the Blender file with which the cockpit lightmaps were created ? The two in the dev folder don't contain any light sources that could be used for lightmap baking ! :(

TheFGFSEagle:
Hmm, yeah, I guess I can copy the main lightmap and crop it to the right position and size.

TheFGFSEagle:
I think we can spare the extra lightmap - @hbeni, please see if it looks good enough now, and if it does please merge and close this pull request ! 😃

hbeni:
Thank you Eagle!

PR #501, opened 2023-04-24
Fix Clock, Hobbs meter and Yoke lighting problems with latest next

hbeni:


hbeni:
Fixes are obsolete, cause is regression bug in fgfs core/simgear.

PR #507, opened 2023-09-04
Independent back and side glass rain effect

wlbragg:
@hbeni 

Still to fix...
Front window needs animated drops when engine not running
Back window needs animated drops when engine not running
Right side glass to match left side glass better when engine not running.

Please test and provide feedback.


wlbragg:
Something else I don't want to forget. I finally got a good .ac. But to get it I had to import the .ac edit it and export it. I did not use the Blend file in the dev folder. I saved my new modified Blend file but did not push it to the dev repository. I didn't want to overwrite your original. I could rename mine and push it. Or whoever manages your original Blend file can separate the back interior glass from the front and name it Windscreen.Int.Back so you original Blend is up to date.

wlbragg:
@hbeni

I'm done with this. This is about as good as I can get it without spending a bunch more time on it. I also included a Blend file that has the cut back window. It's with the other original blend. @HHS81 can combine or do what needs to be done with it.

hbeni:
Will test as soon as I git time.
Thank you very much!

TheFGFSEagle:
git time … yeah, that would be cool if you could just git commit "2 hours of free time" and have 2 free hours ! 🤣

hbeni:
First quick test looks good. :+1: 

What I would still work on would be this:
- The side of the windshield doesn't move at all; I think it would look better if we could animate that too
- Same for the side windows, they are still, however in flight (or heavy wind) should be moved
- The side windows have too much rain I think. When standing still, there are the wings above them, so they should not have much droplets at all; only when
  -  prop downwash
  - or wind is coming from the front
  should droplets occur.

TheFGFSEagle:
Hmmm, if you want to have no drops under the wings, I think you need to use a custom rain-norm property calculated as `-(splash-vector-x) * tuning-factor * /environment/rain-norm` - then in the effect file set the <rain-norm> to that property.

hbeni:
Yeah, but _where_ in the effect file, where is the parameter for that?

wlbragg:
>The side of the windshield doesn't move at all; I think it would look better if we could animate that too
>Same for the side windows, they are still, however in flight (or heavy wind) should be moved

Unfortunately, you really can't, at least not the same as you can the horizontal surfaces. All you can do with the side windows is to have the drops that hit the window elongated, and if you stretch them enough it kind of gives the appearance that they are moving. What really is happening is the new drops are relocating to a different position. But when you crank the vectors high enough you mostly get vertical displacement, not horizontal displacement and you back to drops that have the appearance of moving vertically, not horizontally. At least I wasn't able to and I used some helper code that let me adjust each vector in real time trying to see what it was capable of.

>Edit: Tried to play with the filters, no luck. Where do i need to adjust this?

Your more than welcome to try to adjust it. There is a couple things you need to know. Read about the side windows, per the [troubleshooting documentation](https://wiki.flightgear.org/ALS_technical_notes#Troubleshooting)

All the vectors are laid out like this
~~~
<!-- front windshield -->
<filter>
        <name>splash-xa</name>
        <update-interval-secs type="double">0.1</update-interval-secs>
        <type>gain</type>
        <gain>1.0</gain>
        <input>
            <expression>
                <table>
                    <property>/velocities/airspeed-kt</property>
                    <entry><ind>  20 </ind><dep>-1.7</dep></entry>
                    <entry><ind>  50 </ind><dep>-1.8</dep></entry>
                </table>
            </expression>
        </input>
        <output>
            <property>/environment/aircraft-effects/splash-xa</property>
        </output>
    </filter>
---
---
<name>splash-ya</name>
---
---
<name>splash-za</name>
---
---
<name>splash-xr</name>
---
---
<name>splash-yr</name>
---
---
<name>splash-za</name>
---
---
~~~

and then the actual filters that decides when to use what vector

~~~
<filter>
        <name>splash-x</name>
        <update-interval-secs type="double">0.1</update-interval-secs>
        <type>gain</type>
        <gain>1.0</gain>
        <input>
            <condition>
                <greater-than-equals>
                    <property>/velocities/airspeed-kt</property>
                    <value>20</value>
                </greater-than-equals>
            </condition>
            <property>/environment/aircraft-effects/splash-xa</property>
        </input>
        <input>
            <condition>
                <less-than>
                    <property>/velocities/airspeed-kt</property>
                    <value>20</value>
                </less-than>
            </condition>
            <property>/environment/aircraft-effects/splash-xr</property>
        </input>
        <output>
            <property>/environment/aircraft-effects/splash-vector-x</property>
        </output>
    </filter>
---
---
<property>/environment/aircraft-effects/splash-vector-y</property>
---
---
<property>/environment/aircraft-effects/splash-vector-z</property>
~~~

I have commented at the start of each set of splash vectors which one it is for.
~~~
<!-- front windshield -->
<!-- left side windows -->
<!--right side windows -->
<!-- back window -->
~~~

There are up to 6 vectors per the 4 different windows above, ie; front, ft, right and back.
splash-xa
splash-ya
splash-za
"a" is for airspeed

splash-xr
splash-yr
splash-zr
"r" is for RPM

I condition on RPM first, all the "r" vectors are influencing the rain effect. Once airspeed reached 20 mph then it is the "a" vectors that are influencing the rain splashes.

These are the interpolated adjustments for each vector, in this case at 20 mph airspeed the vector value is to be -1.7.
At 50 mph the vector value should be -1.8. These are the values you would change.
~~~
<property>/velocities/airspeed-kt</property>
<entry><ind>  20 </ind><dep>-1.7</dep></entry>
<entry><ind>  50 </ind><dep>-1.8</dep></entry>
~~~

Doing it this way we have complete control. We can make as many interpolation entries as we want. This is where the pain to gain comes in. I experimented with this a long time ago and just this recent go at it. We could spend days trying to tune this and probably get it better, but at what cost time wise and how much better?

The easiest way to see what the values actually do to the animation is to make the vector value input dynamic so you can see changes in real time.

What I did to learn more about what these vector do was to do the following for all the vectors, then in the property tree adjust them one at a time.
~~~
<filter>
        <name>splash-x</name>
        <update-interval-secs type="double">0.1</update-interval-secs>
        <type>gain</type>
        <gain>1.0</gain>
<!--
        <input> 
            <condition>
                <greater-than-equals>
                    <property>/velocities/airspeed-kt</property>
                    <value>20</value>
                </greater-than-equals>
            </condition>
            <property>/environment/aircraft-effects/splash-xa</property>
        </input>
        <input>
            <condition>
                <less-than>
                    <property>/velocities/airspeed-kt</property>
                    <value>20</value>
                </less-than>
            </condition>
            <property>/environment/aircraft-effects/splash-xr</property>
        </input>
-->
<input>
      <property>a-vector-x</property>
</input>
        <output>
            <property>/environment/aircraft-effects/splash-vector-x</property>
        </output>
    </filter>
~~~

Do that for each splash-vector, x, y and z of each window. Because each of the four different glass surfaces need different vectors. Then you can open the property browser and right at the top of the root you can enter values to see exactly what it does.

hbeni:
Thank you, for this exhaustive explanation ☺️

However, one question still renains: how Van i adjust how *much* rain is rendered on the side windows?
I would want to add a filter that makes sparse droplets with no wind/downwash and more with it

wlbragg:
That exhaustive look was the only way to explain it.

OK, that would be the left side window and right side window sections. That is the tricky part. They don't work as you would think. And they certainly don't work anywhere near as intuitively as the horizontal plane of glass (front an back windows.

I am going to explain the back window so you have a grasp on the mechanics. Here are the back vectors for the airspeed portion.
~~~
<name>splash-back-xa</name>
---
<property>/velocities/airspeed-kt</property>
<entry><ind>  20 </ind><dep> 1.1</dep></entry>
<entry><ind>  50 </ind><dep> 1.3</dep></entry>

<name>splash-back-ya</name>
---
<property>/velocities/airspeed-kt</property>
<entry><ind>  20 </ind><dep> 0.0</dep></entry>
<entry><ind>  50 </ind><dep> 0.0</dep></entry>

<name>splash-back-za</name>
---
<property>/velocities/airspeed-kt</property>
<entry><ind>  20 </ind><dep>-1.3</dep></entry>
<entry><ind>  50 </ind><dep>-1.5</dep></entry>
~~~

At 20 kts or lower the vectors are
x = 1.1
y = 0.0
z = -1.5

If you haven't modeled on Blender this may not make sense to you.
In Blender the back of the model on the X plane is positive, the front is negative. So the vector the rain needs to be traveling on the X, is to the positive if you want it to go from front to back.
So the X value is 1.1
The docs say up to 1 is a stationary drop, to make it move it needs to be higher than 1, thus 1.1. If you set it to 1.5 it will move faster.
The Y plane in Blender is positive to the right, negative to the left. This rain is coming from the front because we are moving forward into it at 20 kt. So no Y vector.
The Z plane in Blender is positive above to negative below. So to get the drop to move down (back window it needs to move down, the front window it needs to move up.) you need a negative number. Again less than 1 = stationary drop. Greater than 1 = the drop starts to move. So the Z value is -1.5.

Now this will be the same concept for all the windows, each using their own vectors.
That is the basics. Each direction or plane of glass will need different values in different directions. 

Now the right side windows. And lets only look at the airspeed portion of the logic again to help explain it.
~~~
<name>splash-right-xa</name>
---
<property>/velocities/airspeed-kt</property>
<entry><ind>  20 </ind><dep>-0.4</dep></entry>
<entry><ind>  50 </ind><dep>-0.5</dep></entry>

<name>splash-right-ya</name>
---
<property>/velocities/airspeed-kt</property>
<entry><ind>  20 </ind><dep>-0.2</dep></entry>
<entry><ind>  50 </ind><dep>-0.3</dep></entry>

<name>splash-right-za</name>
---
<property>/velocities/airspeed-kt</property>
<entry><ind>  20 </ind><dep>0.4</dep></entry>
<entry><ind>  50 </ind><dep>0.5</dep></entry>
~~~

At 20 ks the values are...
x = -0.4
y = -0.2
z = 0.4

Ok, you want less drops at 20 kts or lower, so those values need to be reduced. But too little and then you get no new drops animating at all.
Maybe...
x = -0.3
y = -0.1
z = 0.3

But herein lies the difficult part. The side or vertical glass does not function at all like the horizontal glass
First it needs the Y vector. Remember the right glass needs a negative Y value (from right = pos to left = neg). So to hit the outside right glass you need a negative Y. We have -0.2
The rain on the vertical glass does not animate the same as on the horizontal glass. You really only change the circle drop into an oval drop by increasing the vector values. To a point where it is a very long, sharp oval. If you keep going in value it starts replacing those sharp ovals with new ones that gives you the illusion of rain moving. But then it acts weird and instead of going from front to back it always goes from bottom to top. In other words the first frame places long thin ovals horizontally in the center of the window, the next frame removes that oval and places the next one higher than the first one. If you can figure out the exact amount of x-y-z vector values you might get the next oval to be placed farther towards the back of the window than the first one. But it never really looks like rain drops moving from front to back like the front and back horizontal glas does.

The X you would think should be a positive value if it is acting like the horizontal glass did, right. Notice it is a negative, -0.3. That is because a positive probably didn't give me a drop at all or it did something really strange. Notice the Z is a positive vector. you would think it maybe should be a 0 right. Not o, again you might not get any drops or any new animated drops if you don't give it a value. The side windows are really difficult., I would need to spend a lot more time on them to make any significant changes. That is why I stopped where it is now.

Hopefully this explains it a little better. I can try to use my testing method to see if I can get the sides to look better. I think less drops is possible, but moving drops are really problematic. And less drops suddenly mean NO drops. But try it, maybe you will have better luck than I did.

After explaining it again, I feel like I want to try again.

hbeni:
Thanks again...

However what i not get is *why* the side windows behave differently, than the front windows. Is this because they are in other orientation?

For less drops: I think a good compromise would be to have no drops when the rain is falling down straight, and to increase drop count (if that is possible) with increasing air speed/wind/downwash.
I played with the `/environment/rain-norm` [Eagle mentioned](https://github.com/HHS81/c182s/pull/507#issuecomment-1708955227), but found no way to tell the effect to use another property calculated by me.

TheFGFSEagle:
Uh, in your effect derived from model-combined-transparent (or glass ?) applied to the windows, in the <parameters> section you might have (if not, just add it):
```xml
<rain-norm><use>/environment/rain-norm</use></rain-norm>
```
Just change the property to whatever you want ! :) If you use a custom property however, make sure to transmit it over MP and use a property path without leading / (this might make for awkward situations however - like the pilot having a rainy weather but another MP participant having sun, and the rain drops being still rendered :mrgreen: )

TheFGFSEagle:
That problem could be solved by transmitting a factor for rain-norm and have a Nasal timer in the <nasal><load> of your model XML that multiplies the local /environment/rain-norm with the factor, and on the main instance you'd have a property-rule for that.

hbeni:
@TheFGFSEagle The parameter is `<rnorm>`, otherwise, it's what i want

hbeni:
Scaling does'nt look good. We need discrete values.

TheFGFSEagle:
Whoops, yesh, rnorm it was, sorry for the confusion.

hbeni:
added d3badb9 and this reduces the rain on the side windows.
I separated the sides effects from the sides of the front windshield, so we can adjust independently.

What do you think?

hbeni:
> The rain on the vertical glass does not animate the same as on the horizontal glass.

sorry to bring it up, but that piece still puzzles me.
Why is the animation different? And how can we change that?
I think the side parts of the windshield would benefit greatly from this insight.

wlbragg:
>I separated the sides effects from the sides of the front windshield, so we can adjust independently.

Yes I think that is a good idea. I think the 172 are separate, but I include them as a side window group.

>Why is the animation different? And how can we change that?

I really think it is a shader mistake, you can't get it to go down, why? I'll see if I can find it and make sense of it.

TheFGFSEagle:
Uh, if you flip the sign of the Z component it should go down just fine ?

hbeni:
No. As wlbragg explained its unfortunately not nearly as easy

hbeni:
And i would be perfectly happy if the front sides would behave like the front, that is not accurate but the immersion is better

wlbragg:
That's why I think the shader might be incorrect for the z on the vertical.

wlbragg:
I guess I should quit saying vertical and horizontal. More accurately it is on the x plane VS on the y plane. I guess you'd have to stand the front grass straight up and see if you can still get z movement up and down to truly know if they are different.

TheFGFSEagle:
So you cannot get any downward movement ? Not on the  172 either ? Inverting the normals doesn't help ?

wlbragg:
How do you invert the normals? If you invert in Blender and then export, you no longer see the effect. Can you only invert the Z normals? That might actually work.

wlbragg:
As far as I know the glass is one sided and the normal point in.

TheFGFSEagle:
One more thing you could try is UV-mapping your windows and using `<surface-mapping-scheme>1</surface-mapping-scheme>`

TheFGFSEagle:
[The wiki](https://wiki.flightgear.org/ALS_technical_notes#Troubleshooting) says:
> […] in particular for near vertical cockpit side windows the scheme does poorly.
> 
> …
>
> If the parameter is changed to 1, the uv-mapping of the surface is used (it has to exist of course). For this to work properly, the uv-mapping needs to be sufficiently regular and roughly preserve the mapped area.
> 
> If the parameter is changed to 2, a local orthonormal system based on the normals of the surface is constructed. This no longer takes the splash vector consistently into account and works poorly for surfaces which are flat in the (xy)-plane, but gives decent result for side windows.

TheFGFSEagle:
> I had to visualize it in Blender. I wasn't thinking straight. I was thinking if it had thickness then you could flip the edge normals (Z)

Yeah, FG doesn't even know about edge and vertex normals when it is fed with an AC file ! ;)

wlbragg:
OK, I think I understand now what that means, I read it but it didn't sink in at the time. I'm pretty sure both the 182 and 172 are properly mapped. But i will look at that again in context.

wlbragg:
Son of a B, I tilt that side glass and we can get the down direction. Now, how little of a tilt do we have to have.

wlbragg:
Nope, it requires close to 20-30% tilt to make it behave as expected. I tried rotating the uv-mapping, I tried mirroring the mapping. Nothing fixes this. I am back to if the shader can't the perpendicular nature and errs to up. Why can't we at least make it error to down?

wlbragg:
@hbeni 

The reason I originally left the effect where I did, before the novel of new information and details, is because of the current issue. After a whole other round of testing and expanding my understanding of what is going on, I am left in the exact same position. I think the best we can get is to simulate less drops, as you suggested, and that is it. Can the front and back windows be tuned a little better, yes probably. The rain vector addon would certainly help with that. But I'm afraid without some shader intervention we are stuck with what we have now barring any minor adjustments. Unless @TheFGFSEagle comes to the rescue once again and figures something out, ha!

wlbragg:
Well, I think I was correct. @TheFGFSEagle check out the shader glass-ALS.vert

Change
vec3 base_1 = vec3 (-corrected_splash.y, corrected_splash.x, 0.0);
to 
vec3 base_1 = vec3 (corrected_splash.y, corrected_splash.x, 0.0);

and the drops flow down instead of up. If it i changing nothing else, even if it reversed the front and back flow, who cares so long as we can edit front and back directions, We can't the side so they need to be hardcoded, or like I said "errored" down.

I guess I need to post this to the mailing list and ask the question. But I was hoping for some input here first.

wlbragg:
Wow, it didn't even change the direction on the front and back windows.

TheFGFSEagle:
welll … the problem is, now the drops can't go up anymore ! XD Which might be desirable in some situations e,g, high alpha. But I guess, better than nothing.

TheFGFSEagle:
I'll have a look at that shader when I get time.

wlbragg:
>Which might be desirable in some situations e,g, high alpha

99% of the time we will want them to go down VS 1% we want them to go up. No argument as far as I am concerned. The more I look at that shader line, I really don't understand why the sign is changed there?

wlbragg:
@hbeni 

Here is the latest test I did with the c172p. This is with all side glass modeled as front glass in the .ac file and then using a material animation to place them in the side positions where they belong. It is using scheme 1. The entire side glass, front, windows and back sides is all using the same splash vectors. I still need to possibly use your side window splash vector definitions on our side windows and back side windows. I would leave the front side glass (windshield) with the vectors shown in this video. You can see a small portion of the left side window glass. I would most likely slow that down or remove all animation from it like you have it.

wlbragg:
@hbeni 

See https://github.com/c172p-team/c172p/pull/1453#issuecomment-1719884398

I think I have the effect as good as it is going to get. There is a limit to how little of the effect you can put on the glass under the wings. I tuned it to the least amount I could. I am actually pretty happy with the final results all around. Perfect, no. The best were probably going to be able to get, I really think so. I am not sure where your at with the 182. I ended up using the hack of moving the glass in the blend to a front facing glass and then rotating in the xml. 
There was one questionable effect of when at airspeed or prop speed whether to leave both the raindrop splash animation and the rain moving on the glass. It could be argued to just do the moving drop an not the splashes. I chose to use both on the surfaces exposed to rain but only the moving drops on the glass under the windows. You don't ever see drops splashing on the windows covered by the wings. 

The glitches with unstable rpm values were worked around by limiting how much interpolation was allowed around the unstable ranges. The effects are tuned pretty well around these instabilities so they stay consistent and stable on rpm and airspeed changes.

hbeni:
> If you please, take a look at the final effect and let me know if there is anything you need on the 182 side.

The video looks very nice :)

I adjusted the front a little, and I think this is good-enough now.
The side windows, I am happy with not moving droplets; when this only can be changed by rotating the glass in the AC I think @HHS81 would have objections.

hbeni:
Would be nice to get some testers.
Just start in "core low pressure region" weather and play with the engine/plane/wind.

PR #508, opened 2023-09-07
Rain editor

wlbragg:
This is for testing the rain vector editor only. Delete when finished with it.

DO NOT MERGE!

@hbeni 
I made this real quick to help visualize the rain on the glass. It is a little slider menu to adjust the rain splash on every window.
To use just watch the window you want to get the numbers for and adjust the sliders for the best effet. Use that number where needed in the interpolation table.

I put the button to open this GUI at the bottom of Aircraft Options .

I didn't know if everyone would have access to it without making a PR?

After using this and trying all the surface-mapping-scheme's, I don't know if there are 0, 1 and 2 or what? I really didn't notice much of a difference in them. But I did notice for sure you can only animate the drops going up. I think maybe someone messed up the math in the shader for the side (vertical) glass. I'd be willing to bet a simple sign change and this works the way we want it to. I'll have to take a look at the shader.


hbeni:
Thx will mich lay with it ☺️

hbeni:
OK. I now see what you mean.
That is dancing on a ball, juggling with knives.

Still, i wonder if there is a way to simply tune down the rain.
I currently  play with eagles comment on that.

hbeni:
I close this PR; the branch can remain here for now.
Nice tool btw :)

wlbragg:
@hbeni I pushed this rain vector helper to Addons as a addon. It will work with the c182 and 172 as is. You just need to comment out the autopilot call to rain-effect.xml so it doesn't control the effect anymore when you use it. That way you don't need the duplicate c182 just to use the helper.

TheFGFSEagle:
Heh, great ! :D

Issue #510, opened 2023-10-08
Garmin GPSMap 196 on yoke


jakkos98:
Hi
I managed to add Garmin GPSMap 196 from fgdata to C182s.
Since my goal was to add this GPS in such a way that it would not interfere with the observation above the glareshield and without modifying the panel model, I decided to attach it to the pilot's yoke.

The current implementation allows the GPS to be hidden with the same level of transparency as the yoke when hidden, but lacks the GUI or clickspot to completely remove it from the cockpit as in the C172p when hidden in the glovebox and correct backlight.
![image](https://github.com/HHS81/c182s/assets/40403643/97188f46-c95b-4f1a-81c9-687c943872de)
![image](https://github.com/HHS81/c182s/assets/40403643/51fd7426-481c-47c8-87e4-7313fe594b9c)
![image](https://github.com/HHS81/c182s/assets/40403643/16806fa9-0a60-4882-8d2a-3ba70a062cf0)

https://github.com/HHS81/c182s/assets/40403643/59c44808-4a3e-4cff-a4b0-aa075724a0d8

To add this GPS you need to modify/add the following files:

- c182s-set from main aircraft directory 
- interior-model-c182s in c182s/Models
- copy garmin196 directory from fgddata to c182s/Models/Instruments
- optionaly garmin196 texture from c172s if prefered black insted of grey

Bellow I attach those files, but I'm unsure if are compatible with latest c182s version, so I advice to copy only parts of code needed to garmin 196 (look for garmin196 in code).
[c182s_garmin196.zip](https://github.com/HHS81/c182s/files/12840335/c182s_garmin196.zip)

Best regards
Pozdrawiam

Jakub Kosakowski

hbeni:
Cool addition, but I have some questions:
- a good thing would be if you cloned the repo and filed a pull request containing the changes
- How is this affecting FPS/Performance?
- How exactly is the GPS integrated into the planes avionics (failure simulation, electric power, gps circuit braker?)
- I think a good idea would be to add a (persistent) checkbox in the aircraft dialog, and when switched off completely removes the gps. Would be good id this would save all performance affecting calculations.

jakkos98:
1. I will clone the repository and submit a pull request.
2. I don't see any significant impact on FPS. As with C172, slight delay when starting the unit (can be removed by disabling city list loading).
[...]

PR #511, opened 2023-10-08
GPSMap196

jakkos98:
#510
Initial Garmin GPSMap196 implementation from fgdata for C182S 

To do:

- [x] Add checkbox in aircraft dialog for enabling/removing unit from cockpit
- [ ] Refere to gpsmap196 from fgdata instead coping all files to c182s 
- [ ] Add appropriate night textures and lighting
- [ ] (Optionally) Create a 3D model of GPS yoke mount https://www.garmin.com/en-US/p/6709 (need help as I have no experiene in 3D modeling)
- [ ] (Optionally) Add selections of different locations where GPS can be placed such as on glareshield, built-in panel, etc.


TheFGFSEagle:
For the 3D model of the mounting thing, do you have an image of it being mounted on the yoke ? I can't quite imagine where it would mount to on the yoke ! XD

jakkos98:
https://youtu.be/s0l6AQ5xprI?feature=shared

Same principle as with this iPad mount.

hbeni:
Another question: If the instrument is already in FGAddon, why do we need to copy it over to this plane instead of just referencing it?

jakkos98:
Good point it can be referred to one from fgdata. I simply forgot about it, because when I was firstly tried to add it to another planes few year ago I was using standalone one from FG PAF forums.

hbeni:
> Good point it can be referred to one from fgdata. I simply forgot about it, because when I was firstly tried to add it to another planes few year ago I was using standalone one from FG PAF forums.

That would be a very good thing indeed, because this way we benefit from bugfixes and imporvements without the hassle to track and port this ourselves

hbeni:
@jakkos98 Hi, are you still on this? :)

jakkos98:
> @jakkos98 Hi, are you still on this? :)

This is still on my to-do list. Unfortunately, I haven't had much free time in recent weeks.

PR #523, opened 2023-11-13
Compass enhancements

hbeni:
- Compass deviation now is properly influenced by electrical power
- Compass now models swinging on a pivot pin
- Compass sticks when out of the pivots stick limit


hbeni:
There is an ongoing discussion wether the dip error (pitch error) sitting at the apron is wrong.
If so we could model that by the pitch-offset we already have an filter for :)

in my opinion this is wrong (sources i have seen only describe the error when turning or when accellerating/decelerating; as does a code comment in flightgears instrument code, too!), but i have no real compass experience.
Maybe we can get something from the compass manual.

hbeni:
> actually… if in doubt i‘m probably the one who is wrong

OTOH, POH states that the DG should be set directly prior to takeoff against the compass ... which is currently not possible due to errors on a W/E heading. So, I'm probably not so wrong.
Also the compass card is hung on a pivot needle, which helps to counter the Dip error effect.
So unless the compass hits some pitch limit, it should move freely and show no error on W/E headings, unless acceleration changes.
Or?

hbeni:
I added a simple pivot mounting which will set a lagging offset (so pitch changes still influence the compass) for aircraft pitch -9° to +9°.
This will let the compass be a reference for calibrating the DG before takeoff when standing still and within limits.
In air the pivot will also move freely but honor for example rapid dives.

hbeni:
From the mailng list:
> Hi
>
>I did find some information.
>1. Yes the card has a hub that sits on a needle and it not only rotates, but also swings (roll and pitch) some degrees. (9° may be good)
>2. x acceleration lets it pitch, y acceleration lets it roll. (BTW the code for computing the angle  4 degrees (0.07 radians) for every G (32 fps/sec) of acceleration. is wrong. one g would be enough for 45°)
Standing at runway short to departure is the same as steady flight -> dial parallel to horizon
>3. The plane's pitch angle plus the dial's pitch angle, and the plane's roll angle plus the dial's roll angle are relevant for the orientation towards the local magnetic field, especially the inclination.
>4. The formulas are from the link in the source. one has to use archive.org, then click up , up and pdf. http://williams.best.vwh.net/compass/compass.pdf
>5. Should the swing limit be reached, the edge of the dial touches the case and causes too much friction to let the dial >follow the magnetic field, it simply sticks. When the dials gets free again, it immediately starts to turn to the correct orientation.
>6. When looked in depth, one needs to consider the mass and viscosity of the damping fluid s well, as it starts swirling when the case rotates, and tends to take the dial with it some more degrees..
>I found a report on aircraft compasses being rotated. You may neglect the effect :-)
>
>Why do I know that?
>I am building a mechanical magnetic compass for a C172 home cockpit.
>
>Regards,
>
>Oliver

So the "sticking" should be modelled too IMHO.

PR #524, opened 2023-11-13
Revise NAV needle deflections

### LINKED ISSUE #522: Revise NAV1/NAV2 needle deflection values
The CDI should represent 2° off per dot, this is slightly off currently.

The VOR gauges currently just scale the deflection using a fixed factor, and that is off somewhat.
Full deflection (=10°) should give 5 dots.

- [x] NAV1
- [x] NAV2

----
Ref:
- Manual https://www.manualslib.com/products/Honeywell-Ki-209-8777567.html
-  Wiki https://en.wikipedia.org/wiki/Course_deviation_indicator

hbeni:
Fix #522

### DIFF
--- a/Models/Instruments/vor/vor.xml
+++ b/Models/Instruments/vor/vor.xml
@@ -74,7 +74,19 @@
   <type>rotate</type>
   <object-name>LocalizerNeedle</object-name>
   <property alias="../../params/nav-heading-needle-deflection"/>
-  <factor>-2.8</factor>
+  <interpolation>
+    <entry><ind> -10 </ind><dep>  28 </dep> </entry>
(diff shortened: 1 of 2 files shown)


hbeni:
Tested and ready for merge, imho

wlbragg:
@hbeni do you know if this applies to c172p?

hbeni:
No, but its easy to check; just temporarily link the needle to a test prop which you can adjust using the prop browser.

Each dot should represent 2degrees using a VOR (tune to that, ILS is more sensitive but will automatically match if the VOR/2deg is correct)

wlbragg:
OK, thanks.

hbeni:
Just briefly checked:
- NAV1 is good on the right, and slightly off to the left
- NAV2 is off on both sides

PR #527, opened 2023-11-19
Issue525 save dg offset (by switching instrument to <heading-indicator-dg>)

hbeni:
This switches the heading indicator to the `heading-indicator-dg` implementation which is more powerful and realistic.

:warning: PR is based on switching to the DG based heading indicator, and only works with the patch to the core:
https://sourceforge.net/p/flightgear/codetickets/2839/

:information_source: Regarding the movement-errors, there is discussion on the mailing list if this is realistic:  https://sourceforge.net/p/flightgear/mailman/flightgear-devel/thread/6e66f08e3fb3f02a9d5c557cfd7a2910%40hallinger.org/#msg53245268
My impression is, that there should be no permanent movement related errors, as the gyro should move relatively freely due to the two axis bearing.

regarding to the videos:
- Uncaged maneuver limits are (for air driven): +-55° in pitch or roll (two videos below say this). IMHO the current c++ coded limits are too rigid. And also it would be nice to have a caging feature.
- Mechanical limits prevent completely freely movement.

----
- [x] Switch to new DG
- [x] error value: only let this apply when the limits (55° pitch/bank) are exceeded
- :warning: _Before merge_;
  - [x] Wait for SF ticket [2839](https://sourceforge.net/p/flightgear/codetickets/2839/) being merged to FGFS/next (so `<suction>` is a supported source of the instrument)
  - [x] Watch SF ticket [2840](https://sourceforge.net/p/flightgear/codetickets/2840/) regarding error discussion (Maybe the filter clamping the error prop can be removed)

----

Refs:
- https://www.youtube.com/watch?v=hVsx4XWafXg Gyroscopic Instruments
- https://m.youtube.com/watch?v=9EZ9ZI8kAWM Operation And Errors Of Directional Gyro Indicator | Lecture 27
- https://www.youtube.com/watch?v=q1TVR9xdl64 The Heading Indicator (Directional Gyro)
- Mailing list discussion: https://sourceforge.net/p/flightgear/mailman/flightgear-devel/thread/6e66f08e3fb3f02a9d5c557cfd7a2910%40hallinger.org/#msg53245268
- FGFS Wiki Newsletter Jan/2024: https://wiki.flightgear.org/FlightGear_Newsletter_January_2024#Gyro_based_instruments_enhancements (with more info)
- C172 sister ticket: https://github.com/c172p-team/c172p/pull/1490

PR #528, opened 2023-11-20
Issue525 save dg offset (by storing offset with simple DG implementation)

### LINKED ISSUE #525: DG: save precession drift between sessions
The DG has drift over time. The drift offset is reset when the sim restarts.
It should be periodically offset by turning the knob on the Instrument, But that is never needed when only doing short flights.
So we should init the drift by the value from the last session.

Precession drift is 360deg/day = 0.25deg per minute (3.75deg in 15 min), so quite big.

- - -

Refs
- https://m.youtube.com/watch?v=9EZ9ZI8kAWM construction
- SF Core ticket: https://sourceforge.net/p/flightgear/codetickets/2838/
- Mailing list discussion: https://sourceforge.net/p/flightgear/mailman/flightgear-devel/thread/6e66f08e3fb3f02a9d5c557cfd7a2910%40hallinger.org/#msg53245268

hbeni:
DG/HI: add "align" property, so knob does not rotate with precession

This is based on FGF SF ticket 2838 (https://sourceforge.net/p/flightgear/codetickets/2838/) for the precession to actually apply to the DG.

If applied, should be compatible to PR https://github.com/HHS81/c182s/pull/527
Also, this is downward compatible to older versions, so could be applied also without the SF ticket already.

Fix #525

### DIFF
--- a/Models/Instruments/hi.xml
+++ b/Models/Instruments/hi.xml
@@ -130,7 +130,7 @@
   <animation>
     <type>rotate</type>
     <object-name>OBS-Knob</object-name>
-     <property>instrumentation/heading-indicator/offset-deg</property>
+     <property>instrumentation/heading-indicator/align-deg</property>
 <factor>10</factor>
 		<center>
 			<x-m>0.012842</x-m>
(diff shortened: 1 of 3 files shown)


hbeni:
(see also discussion in #553 regarding opt-in or opt-out for realism)

PR #534, opened 2023-11-25
Refactor wing lights select animation

### LINKED ISSUE #533: Wing light: not showing emmiting face lit (S/T variant affected)
When at night and activating the rear domeligt, you can switch on exterior lights.

The (compositor) light nicely lights up the side of the plane and the apron, however the wing light mount is not bright.

hbeni:
Fix wing cortesy lights not showing illuminated light source face

fix #533


hbeni:
Refactored the lights select animation and confirmed it to work.

However, the face is still not showing.

PR #535, opened 2023-11-25
Issue526 autostart cold weather

hbeni:
- Fixes some small cosmetic issues with the autostart function
- Failed start with "complex engine oeratoins" is expected, but this was not very well documented.
  Added comments and code that clearly points that out.


hbeni:
Tested and ready for merge

Issue #541, opened 2023-11-29
Autostart: Make more configurable


hbeni:
I have seen that the autostart feature is not entirely consistent with the c172.
For example, we remove tiedowns etc.

Maybe its good when the user can tune some aspects of the autostart function to his likings, so that he can choose gradually between "just kick off the engine" and "do everything I need for a quick flight".

- _Autostart_
  - [x] Run preflight checks (minimum fuel, oil, etc)
  - [x] Repair damage
  - [x] Remove securing
  - [x] Preheat engine
  - [x] Turn on Electrics
  - [x] Autolean to RoP
  - [x] Calibrate Heading Indicator
  - [x] more?
- _Autostop_
  - [x] Secure plane on ground
  - [x] Turn off Electrics

wlbragg:
I think the general thought on the subject was that autostart should do everything to enable the aircraft to taxi and then takeoff. In the c172p we even have auto mixture for airports at altitudes and the taxi and landing lights if autostart is attempted at low light. I don't see any reason not to have different customizations for it, but, also, if one is to do partial checklist items, they might as well do them all? Instead of customising autostart, maybe the startup state system would be more fitting for this type of thing? That is kind of what it was designed for. In fact, I think if I were to make a new aircraft from scratch, I would not have a nasal/gui autostart at all, I would do it as a startup state.

hbeni:
Hm, when I last tested the 172, tiedowns were not removed.

This led me to the impresssion that autostart was meant to just start the engine.
The c182 has autolean too, and some other stuff, and actually currently runs the preflight checklist.
It even does lean RoP for you, and adjust throttle for a fast idle or for cruise when in air....

So maybe, we just close this as wontfix and use the time for other development stuff :)

hbeni:
Also some tought about the „ready to fly“ state.
Autostart currently does not preheat the engine, so in cold weather the engine can die shortly after autostart.

This what i tought „states“ are for: you can switch the plane to „ready for takeoff“, and this too will run the checklist _and_ preheat, so you are really able to directly go and fly.

This is all a little inconsistent currently…
But no big deal.

hbeni:
> Instead of customising autostart, maybe the startup state system would be more fitting for this type of thing? That is kind of what it was designed for. 

Yes, its somewhat duplicating.
However an important difference is that startup states can (without nasal magic) only be selected from the launcher, and autostart is available always. Which is handy in my point if view.

[...]

PR #544, opened 2023-12-01
Issue541 refactor autostart GUI and make configurable

### LINKED ISSUE #542: States at runtime: Apply the "ready for takeoff" state in cold weather at runtime and you end up with a not-running engine
> Also checked: Apply the "ready for takeoff" state in cold weather at runtime and you end up with a not-running engine. That needs to be fixed.

_Originally posted by @hbeni in https://github.com/HHS81/c182s/issues/541#issuecomment-1835605111_

hbeni:
Implements #541
Fix #542 (implicitely, branch is based off that fix)


hbeni:
Tested, I'm excited to see feedback :)

PR #549, opened 2023-12-19
Add Towbar for improved ground handling.

hbeni:
Planes ground handling dialog now gives access to the towbar item. Towbar can also be attached by clickspot on the nosewheel.

When attached, it can be grabbed with the mouse (hold button) and then left/right movement will adjust direction and back/forwad controls force. Once the mouse (=force) is released, the plane will slowly stop.


hbeni:
`Models/Exterior/towbar.ac` is ugly and should be redone by a more capable 3D designer (should not be so much work I guess);
but technically the towbar works quite nicely; with and without walker; and is fun to use :).

- Go to the nosewheel and click it
- In the GUI, "attach the towbar"
- Then grab the bar with the mouse (click+hold) and move the plane around.
  This works mostly intuitive when standing in front towards the plane:
  - left/right movement will change the wheel orientation
  - moving the mouse away from you will ad backwards force, moving the plane away
  - pulling the mouse towards you will pull the plane to your position

When using the walker, you still can use WASD in parallel to get yourself out of the way (or chase your plane, haha :rofl: )
  

grafik

TheFGFSEagle:
How about you copy the tow bar from my  C210 ?

hbeni:
Ah, there is one already?

TheFGFSEagle:
Yes, my P210N has one in Models/p210n.blend

PR #550, opened 2023-12-20
New tow bar model

MariuszXC:
New tow bar.
Currently parts are not joined into one object, to allow for fine tuning. Can be either joined in Blender (which is kind of permanent) or using FG animation (leaves space for future modifications - probably preferred).


hbeni:
Cool, that was quick.

This is how its rendered here:
grafik

Would have expected a more saturated blue and black of the handle and the grips.
And a little bigger, maybe?

grafik


----
Just had a quick shot at scaling:
grafik

But I think it should extend more to behind the prop spinner?

hbeni:
https://www.aopa.org/-/media/Images/AOPA-Main/News-and-Media/Publications/Flight-Training-Magazine/1710f/1710f_pf_tow/1710f_pf_tow_16x9.jpg

hbeni:
Had a shot at it.
Looks good, doesn't it?
grafik

hbeni:
Now only question remaining is, why the color is so off?

MariuszXC:
Dimensions (lenght = 39.5" ) were taken directly from the shop site. Other were eyeballed from the product pic.
Color is off, because you probably need to apply an effect to the object. Actually, the rubber handle ends would need to be excluded from this effect, because rubber is dull, while paint is glossy.. I'll try to move the handle ends to separate objects..

hbeni:
Merged in c284d41

Looks great, @MariuszXC !
Thank you!

PR #552, opened 2023-12-21
Issue 545 towbar

MariuszXC:
Added a long tow bar model (1.4m) .
Accordingly, shortened the previous model to original dimensions (1.0m).

Long version model elements in .ac have "l." prefix prepended to their names, so, for example, "handle" becomes "l.handle" etc.


hbeni:
merged with [e0a5435](https://github.com/HHS81/c182s/commit/e0a543593131ab6d1b16cfd58b13b296c884dbae)

PR #554, opened 2023-12-24
Fix bug that causes FG1000 to not work at all in FlightGear 2020.3

sryu1:
There is no file in FG_ROOT for FG 2020.3 called GMA1357Interface.nas. The file was copied from 2020.4's FG_ROOT to the aircraft directory as a workaround for this.

### DIFF
--- a/Nasal/GMA1347Interface.nas
+++ b/Nasal/GMA1347Interface.nas
@@ -1,0 +1,189 @@
+# Copyright 2021 Stuart Buchanan
+# This file is part of FlightGear.
+#
+# FlightGear is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


hbeni:
Hello and thank you for the Patch.
Unfortinately the master here is I think not entirely compatible with 2020.3 but targeted on next (where your bug is not occurring).
So to fix your issue, the better approach is to upgrade to next (or wait for the release).

We might backport the fix to the current fgaddon stable hangar release however.

sryu1:
Ah ok, thank you and all good 👍

hbeni:
> We might backport the fix to the current fgaddon stable hangar release however.

so just to verify that is indeed good: you tried to use the recent git master together with 2020.3?

sryu1:
Yep

PR #558, opened 2024-01-05
Issue556 convert flashlight to compositor

### LINKED ISSUE #556: Convert flashlight to compositor lights
It may be an idea to convert the flashlight to a compositor light.

This would probably be better because all 3D elements are properly lightened that way, including the plane outside :)

hbeni:
Converts the ALS flashlight to compositor one

- [x] Compositor flashlight
- [x] ~Make usable in walk view to watch plane from outside in the dark (preflight etc!)~
  Needs to be done in [FGData](https://sourceforge.net/p/flightgear/codetickets/2853/)
- [x] ~(optionally) make light direction tied to a mouse vector, instead of fixed view direction~
  Cancelled; its too much hassle for too less gain

Resolves #556


hbeni:
So, its converted and working nicely.
What I really strugle right now is getting the walker x/y/z coords relative to the model one, so to offset the light in conjunction with the current view origin.

TheFGFSEagle:
Replied on the mailing list.

hbeni:
Thanks, will try to define the flashlight for the Walker in the walker model code.
If that works, we can add that to fgdata, wich is a very good idea of you!

hbeni:
Looks good, currently preparing a patch.

hbeni:
OK, this is done so far.
FGData ticket: https://sourceforge.net/p/flightgear/codetickets/2853/

hbeni:
old:
grafik


NEW:
grafik

hbeni:
Added a keybind CTRL+F (which is the proposed one in the FGData patch).
If the patch is accepted, the keybind can be removed too.

hbeni:
The inside flashlight really looks good and is an imporvement.
Will test a little more, but looks done.

Issue #517, opened 2023-10-31
Pitches up with tailwind on ground


dsatgh:
When I'm on the ground with a tailwind somewhere above 10 kts, the plane pitches up until the tail hits the ground. I noticed a big difference in behavior to the c172p. I don't know if it's relevant, but when I look at the wind pitch moment in the properties, the c172p show much smaller values and reacts to elevator movements. The c182s has a much bigger value and doesn't react much to elevator movements.

HHS81:
Did you check that the aircraft AND FlightGear version is the most recent?

I just tried, and yes, the aircraft is pitching up with winds more than 15ktn. The values should have been realistic once I set them, but may be altered some time ago. I'll check that

hbeni:
Also look at the CoG: how was the plane loaded when this occurred? (Tanks, Passengers, Baggage)

dsatgh:
I'm on master branch and put the CoG as much forward as possible.

HHS81:
> Also look at the CoG: how was the plane loaded when this occurred? (Tanks, Passengers, Baggage)

I could reproduce it. The first impression is comparing different numbers, that it is not the elevator- there is something else.

hbeni:
Any new insights into this? :pray:

HHS81:
> Any new insights into this? 🙏

Not yet, at least I can say it is not related with the climb rates issue. I do know that the C182 is sensible to wind on the ground and therefore it will need correct inputs of the yoke wheel in order not to roll over while taxiing, But here there inputs doesn`t have the needed effect. I have to try different things to rule certain things out.

hbeni:
Interesting read, maybe related, on the Forums:

https://forum.flightgear.org/viewtopic.php?p=417356#p417356

> It doesn't actually take much to tip a C172 or PA28 onto its tail; if you go to an airport after a heavy snowfall, you'll find several planes with their nosewheels in the air

and also that the fdm models gear is not spot on the 3D-model ones for the 172, something we should ceck here too

HHS81:
Unless something had been changed on the fdm since I started the model, the fdm gears should be spot on. That was something I checked really early on the model.

hbeni:
Had the mood to poke around with versions a bit to try to get a grip on when this was introduced.
All tests performed with:
* max forward CoG (pilot+copilot full weight, full tanks, otherwise no passengers or baggage) and elevator full down.
[...]

PR #561, opened 2024-01-26
Revert part of commit 7a5a0b6 to fix Issue #517

HHS81:
@hbeni Please check if this solves all the issues, and still feels good- it seems to me that elevator authority is low on approach, but that might be due that I have flown the aircraft not for a long time

### DIFF
--- a/c182s.xml
+++ b/c182s.xml
@@ -1824,10 +1824,11 @@
                     <property>aero/qbarUW-psf</property>
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/cbarw-ft</property>
-
+<sin>
                         <property>aero/alpha-rad</property>
+</sin>
+                    <value>-2</value> 
 
-                    <value>-1.65</value> 
                     <!-- 
                         stall and spin (5):
                         horizontal tail stall (for flat spin)
--- a/c182t.xml
+++ b/c182t.xml
@@ -1823,10 +1823,10 @@
                     <property>aero/qbarUW-psf</property>
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/cbarw-ft</property>
-
+<sin>
                         <property>aero/alpha-rad</property>
-
-                    <value>-1.65</value> 
+</sin>
+                    <value>-2</value> 
                     <!-- 
                         stall and spin (5):
                         horizontal tail stall (for flat spin)


hbeni:
OK; my results:

pitching now happens at abut 65kts tailwind with the [mentioned](https://github.com/HHS81/c182s/issues/517#issuecomment-1910179687) max-forward CoG.
Don't know if this is realistic.

Another thing, the "characteristic" stall break in pitch is not there anymore.

But aside from that, it looks okay, I think.

hbeni:
Maybe something "in between" is achievable, where we have the old effect, but pitch up at say, 30 knots tailwind or so?

HHS81:
> OK; my results:
> 
> pitching now happens at abut 65kts tailwind with the [mentioned](https://github.com/HHS81/c182s/issues/517#issuecomment-1910179687) max-forward CoG. Don't know if this is realistic.
> 
> Another thing, the "characteristic" stall break in pitch is not there anymore.
> 
> But aside from that, it looks okay, I think; especially elevator authotrity at approach looked still reasonable to me.

Can you give me a hint how to reproduce the stall break in pitch? Or do you mean, that the aircraft doesn't stall anymore?

hbeni:
Im no real pilot, but what i Know is that at the moment the stall happens, it should be a small sudden/rapid break down in pitch. This I can't induce with the branch, but without it.

HHS81:
> Im no real pilot, but what i Know is that at the moment the stall happens, it should be a small sudden/rapid break down in pitch. This I can't induce with the branch, but without it.

I see- I just noticed, that the c172p used the same approach as I did- but just the correct coefficients. I'll test

hbeni:
Footage, however not for a 182s, but you see what I mean:
https://www.youtube.com/watch?v=RmWZw71nV98

HHS81:
> Footage, however not for a 182s, but you see what I mean: https://www.youtube.com/watch?v=RmWZw71nV98

Thanks- so you meant Power Off Stalls. It seems the problem behind is bigger: the c172p shows the same behavior by pitching over in tailwind, and stall behavior is less pronounced as I can remember.  I have to dig further in previous versions to understand what is happening

HHS81:
> Footage, however not for a 182s, but you see what I mean: https://www.youtube.com/watch?v=RmWZw71nV98

A much nicer video tutorial: https://www.youtube.com/watch?v=uN6uV_SPYcE

hbeni:
Yes, power off stalls

Hey, Heiko, thanks for taking time to look at this. FDM-side I'm really lost.

PR #566, opened 2024-02-26
Issue564 avionics fan

### LINKED ISSUE #564: Model avionics cooling fan
According to the POH, one preflight checklist item is to verify the avionics cooling fan is audible.

Until the change in fgfs code regarding the gyros, we had some "sound" that was used as proxy (the spinning of the elec gyro of the turn coordinator, which happen to be on the same electrical bus).
Now, the gyros spin-up time is more realistic and thus that sound cannot be used anymore.

We should model the fan:
- [x] Must:
  - [x] add simple simulation of the fan spin-up/spin-down
    - [x] add sound bound to the spin
    - [x] add electrical wiring
- [x] Optional / nice to have:
  - [x] add a simple simulation of avionics temperature
  - [x] model failing of the equipment  due to overheating
  - [x] make the fan failable

----
Ref.:
- Fan is wired at Switch Bus 2 and has a dedicated breaker
- https://megaflow.com.tw/wp-content/uploads/2020/11/4.-Reliability_in_Avionics_Systems_Managing_Excessive_Heat.pdf tells us, that non-critical temps go to about 90°C

### LINKED ISSUE #565: Electrical bus system fixes and code tidying
While looking at #564, I found alot of not correct wirings.
Also the code really needs a cleanup - its bad formatted and thus hard to maintain.

C182T should also be reviewed.

hbeni:
- Add Avionics-fan simulation (including overheating)
- Cleanup electrical system code
- Fix wiring wiring; it's now according the POH
- Battery charge property is now changeable at runtime

**C182t specific**
- Modelled standby battery

---
- Fix #564
- Fix #565


hbeni:
Tested and looking all good

hbeni:
Have seen on the C172, that the GPUs power should only be available with the master switch on.
-> Need to verify if that is the case for the C182 too; but I suspect, it is.

hbeni:
Also, the annunciator panel has power even without power :)

hbeni (commented):
Hi @wlbragg,
Maybe this "combined rocker switch" implementation is of interest for the C172 master switch.
This implementation here has the possibilities:
a) to make the split-behaviour user choosable via an option in the aircraft menu
b) in the combined mode, always both switches are toggled together
c) adds shift-click mode to change the clicked single rocker switch in the "combined- mode"

Currently the C172 rocker switch is a little "erratic" in my opinion:
- with both off, ALT operates both, and BAT operates just BAT
- with both on, BAT operates both, and ALT just ALT

So to switch on both, you need to click ALT, to switch off both, you need to click BAT.
This is "logical" if you follow a flow from left to right to switch on, and right to left to switch off. But it is not documented somewhere, and not obvious from the simulator state (unless you play with it).
Maybe a more consistent approach would be sensible, like "always operate both, regardless which button you click; unless you press shift", and to gibe the user the choice to change to a "single operations mode".

If you like to change the behavior, let me know and I file a PR.

hbeni:
Hi @wlbragg,
Maybe this "combined rocker switch" implementation is of interest for the C172 master switch.
This implementation here has the possibilities:
a) to make the split-behaviour user choosable via an option in the aircraft menu
b) in the combined mode, always both switches are toggled together
c) adds shift-click mode to change the clicked single rocker switch in the "combined- mode"; and "switch both" in "split-mode".

Currently the C172 rocker switch is a little "erratic" in my opinion:
- with both off, ALT operates both, and BAT operates just BAT
- with both on, BAT operates both, and ALT just ALT

So to switch on both, you need to click ALT, to switch off both, you need to click BAT.
This is "logical" if you follow a flow from left to right to switch on, and right to left to switch off. But it is not documented somewhere, and not obvious from the simulator state (unless you play with it).
Maybe a more consistent approach would be sensible, like "always operate both, regardless which button you click; unless you press shift", and to gibe the user the choice to change to a "single operations mode".

If you like to change the behavior, let me know and I file a PR.

TheFGFSEagle:
A little off-topic, I know, but - does anyone know how to add newlines in tooltips ? Are they supported at all ? If I use a `\n` linefeed it is shown as-is, instead of producing a newline.

hbeni:
Eagle, you need to make the tooltip via nasal, there the \n is parsed.
See the tooltip of the master switch in this very pr

TheFGFSEagle:
… or use a literal newline - it's because FG doesn't interpret node text, thanks for the tip ! :D

hbeni:
Long test flight at wednesday was OK

PR #569, opened 2024-03-04
Remove ALS shadow volume

### LINKED ISSUE #568: Remove old shadow volume
Since the addition of the compositor, the old shadow volume is obsolete.
We still have the prop, effect file and the GUI option for that, and that should be removed.

hbeni:
Fix #568


hbeni:
Long test flight at wednesday was OK

PR #571, opened 2024-03-12
Issue 570 lever adjustments for small changes

### LINKED ISSUE #570: Throttle / Prop / Mix: Mousewheel scroll is too aggressive
Currently, the mouse scroll wheel adjustment is somewhat aggressive with the engine controls.

- [x] Maybe setting that from +-5% to +-1% would make adjustments more easy.
- [x] Especially the prop lever is basicly just operable in the upper half, so we might adjust the mapping there too to allow more finer control. That should be backed up by some footage, however.

----
Footage:
- https://www.youtube.com/watch?v=CUCLclB8Nt8&t=350s
  2300RPM in the vid corresponds to about a 80% setting inSim
  grafik

hbeni:
- Add friction lock / vernier feature
- Ammend travel of prop lever

Fix #570

### DIFF
--- a/Documentation.md
+++ b/Documentation.md
@@ -194,6 +194,15 @@ Only passengers inside the plane will jump out, so add them using the "fuel and 
 Jump procedure is that they will crawl outside to the strut, then release; so plan for the drag.
 
 
+Engine lever operation
+----------------------
+The Throttle, Prop and Mixture control levers can be operated either by keyboard,
+by mouse dragging or by mouse wheel (when pointing on the lever and turning the wheel):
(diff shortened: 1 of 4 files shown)


hbeni:
@TheFGFSEagle If you don't mind; would you verify this?

I think its ready and good as it is now.

hbeni:
(Edit: Now this is also in-sync with the c172p: https://github.com/c172p-team/c172p/pull/1498)

TheFGFSEagle:
Will test later this day ! :)

PR #573, opened 2024-03-16
Starter overheating simulation

### LINKED ISSUE #567: Starter motor: Simulate overheated aircraft starter (POH 4-21)
Right now, you can crank till the battery has run down.

The starter should be susceptible for misuse, so one should follow the POH (4-21) for engaging the starter.
This should currently unfold with the overprimed- or underprimed usecase, because otherwise the engine usually will fire.

grafik

----
Ref:
- starter article with some links: https://www.qaa.com/blog/overheated-aircraft-starters

hbeni:
The simulation is very basic, but I think good enough for that usecase.

You can crank the engine about 30 seconds (depending on OAT) before the starter melts.
The POH starter cycle (10s crank, 20s cooldown; then repeatable 2 times, then cooldown for 10 mins) is OK in warm weather (however not anymore in _very_ hot temps!).

It's implemented optional (-> aircraft menu) and defaults to off.

Fix #567

### DIFF
--- a/Documentation.md
+++ b/Documentation.md
@@ -32,6 +32,7 @@ To prevent this, you should apply the plane cover when securing your craft.
 Unchecking `Complex engine procedures and failures` allows a fairly basic engine mode where no priming is needed (just activate
 battery, apply enough mixture and press `s` to start), and no further damage can occur.  
 Otherwise you need to pay attention to varios engine related things:
+
 - Priming for startup  
   The priming procedure involves applying some fuel into the manifold by the fuel pump. The exact procedure is laid out in the checklist.
   The engine can be easily flooded, especially in cold weather (remedy actions also in the POH/Checklist).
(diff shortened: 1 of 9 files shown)


hbeni:
@wlbragg Do you like that for the 172 too?
Ill file a PR then.

wlbragg:
@hbeni I just now read this. I have a couple questions. How bad does this melt in 30 seconds under the worst conditions (really hot day)? What kind of damage will the simulation be, as in it quits working, or needs to cool down or what?

My only concern with something like this is real life accuracy in the final effect. Not how we program it to achieve the end result, but the final effect. I don't know that any appreciable damage will occur if you crank that long on a hot day. It may be more of a theoretical issue, thus they add it to the POH. It may fry one in 1000 starters, something like that. I just wouldn't want to be overly sensitive to damage unless it is real life possibility. I'm considering any damage modeling we have done to these models, like the wing damage and the gear damage. I don't know how accurate or "sensitive" those simulations are either. But the short answer is, yes, would like any "legit" damage modeling we come up with included in the 172. I need to review the failure system for the 172 and make sure it is up to date and robust.

hbeni:
Hi there, the behaviour is easy to tune down if needed.
Currently it is somewhat sensitive in that not observing the crank duty cycle will make the starter inoperable - it just wont do anything.
The „Repair“ button will fix it (simulates a replacement from the workshop).

I did not found real world numbers for how long you can crank till it dies, unfortunately.
I found a lycoming note that they receive periodically dead engines, where the starter was too hot and dying, and that following the duty cycle is important.

Probably the simulation is too sensitive, but I tought this to be a good thing as it teaches correct behaviour and to observe the POH (and The simulation is also opt-in and not active by default.

TheFGFSEagle:
I think this could be helpful: https://www.theaemt.com/resource/excessive-heat-in-electric-motors-a-common-root-cause-of-insulation-failure.html#:~:text=Maintenance%20experts%20agree%20that%20excessive,life%20is%20cut%20in%20half.

> Maintenance experts agree that excessive heat will cause rapid deterioration of the winding insulation within motors. The common rule states that, for every 10°C of additional heat to the windings, motor insulation life is cut in half.

Once that insulation is gone, you'll get a nice short-circuit - at which point either the breaker kicks in, or (in the rare case (does it even exist ?)) that that isn't working as well, a starter fire ! :)

https://www.quora.com/Can-a-starter-catch-on-fire

wlbragg:
OK, if it's not too late and you get the time to do it, I would include it in the 172. I can always grab the code myself an port it if needed.

hbeni:
> OK, if it's not too late and you get the time to do it, I would include it in the 172. I can always grab the code myself an port it if needed.

See: https://github.com/c172p-team/c172p/pull/1506

PR #576, opened 2024-03-27
Rudder: made more unsensitive

hbeni:
Adjust rudder effect:
- On slow speeds, the rudder now does not have a base effect.
  This essentially makes the nose wheel the only effective force source for steering, unless the rudder gets enough wind to be effective. Before this, the rudder always had some effect.
- Added a damping kinematic for rudder traversal.
  This is not significant or big, but helps filter out jitter/fast movements of analog rudder pedals (aka joystick)

As a result, the rudder gets more mushy on very low speeds and it thus takes more right rudder for takeoff.
But I think the right-rudder that was needed before was a) too sensitive and b) too few.

### DIFF
--- a/c182s.xml
+++ b/c182s.xml
@@ -2108,6 +2108,7 @@
                     <table>
                         <independentVar>velocities/u-fps</independentVar>
                         <tableData>
+                                0.0   0.0000
                               107.1  -0.0604 	 <!--Approach-->
                               133.5  -0.0645 <!--climb-->
                               220.1   -0.0805 	 <!--cruise-->
--- a/c182s.xml
+++ b/c182s.xml
@@ -943,8 +943,22 @@
   </channel>
   
         <channel name="Yaw">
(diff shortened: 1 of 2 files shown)


hbeni:
Feedback / Testing appreciated!

ysopflying:
Cndr 0 is not correct. Zero effectiveness are zero speed is already considered in qbar.

TheFGFSEagle:
> Cndr 0 is not correct. Zero effectiveness are zero speed is already considered in qbar.

Yeah, that was what I was saying about the elevator in the other thread - thanks for confirming ! :D

hbeni:
Merged ee6e31ed4c5958674 into master (smalllll mechanical damping of rudder input).
Closing the rest of the PR as discussion results above indicate.

Octal450:
This, is bad.
Not the proper way to do this! It should be done aerodynamically!!!

See my comment in #582 

Kind Regards,
Josh

PR #579, opened 2024-03-30
Added static friction on low RPM

### LINKED ISSUE #578: Prop continues to spin very slowly after cut-off (without complex engine simulation)
User reported Prop spinning very slowly after cut-off: https://forum.flightgear.org/viewtopic.php?p=420372#p420372

This also prevents the towbar from being attachable.

> I noticed that I could not attach the towbar. Then I noticed that after engine shut down the rpm stays between 0.06 and 0.09. I think the towbar wants it to be less than 0.01 rpm.
> Wind: none
>International Standard Atmosphere
It looks like only the C182's propeller doesn't want to stop completely. C172 and DA40NG stop fully

- - -
(Maybe related to #303, but probably not)

hbeni:
The complex engine simulation has some friction caluclation based on oil temps and viscosity, always getting the prop to a full halt in time. Without the complex simulation there was no friction at all. With a dead engine there needs to be some (internal engine) friction to make the prop also stop in time.

fix #578

### DIFF
--- a/Systems/propulsion.xml
+++ b/Systems/propulsion.xml
@@ -692,6 +692,14 @@
             
     -->
     <channel name="Engine Friction" execrate="8">
+        <!-- Add friction with very low RPM, so prop can finally be stopped (Issue #578) -->
+        <switch name="systems/propulsion/static-friction">
+            <default value="0"/>
+            <test logic="AND" value="1.5">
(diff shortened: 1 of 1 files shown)


hbeni:
Positive user testing feedback, merging

PR #583, opened 2024-04-03
Ported Octal450 input filtering from PA28

### LINKED ISSUE #575: Rudder pedals very sensitive
The pedals are much more sensitive than with the 172...
Should that be the case?

Especially with taxiing this is sometimes problematic.

hbeni:
fix #575

### DIFF
--- a/Models/interior-model-c182s.xml
+++ b/Models/interior-model-c182s.xml
@@ -1092,9 +1092,23 @@
         </center>
         <action>
             <binding>
+                <condition>
+                    <not><property>/devices/status/keyboard/alt</property></not>
+                </condition>
                 <command>property-adjust</command>
                 <property>controls/flight/elevator-trim</property>
                 <factor>0.01</factor>
+                <min>-1</min>
+                <max>1</max>
+                <wrap>0</wrap>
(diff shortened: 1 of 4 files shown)


hbeni:
Here is another try. Now It has the lag filters Josh proposed.

Flight test was good.

Two (unrelated) things I noted:
- [ ] The rudder trim wheel looks like its rotating backwards. clockwise rotation (face moves to the left) should dial in left rudder, I think?
- [ ] Both trim wheels are rather coarse when operated with the mouse. The C172 adjusts the rudder by 0.001 steps per click and 0.01 in shifted mode. I think we should do the same. We adjust by 0.01 and 0.1 respective.

hbeni:
I added the "ALT" mode recently introduced at the engine levers.
With that you can now fine tune.

hbeni:
> Please make separate commits for separate changes, or at least list all changes in the commit description - it greatly facilitates it later when someone reads through the commit history !

done

TheFGFSEagle:
Thanks ! 👍🏼

hbeni:
So, any objections against a merge?

PR #585, opened 2024-04-08
Compass correction card: Zoom via canvas dialog

hbeni:
Adding a "zoom function" to the compass correction card.

Clicking it will show a canvas dialog with a better readable card, so the pilot does not need to zoom into the card manually (wich I feel is tedious).

grafik

### DIFF
--- a/Models/Instruments/CompassCorrectionCard.xml
+++ b/Models/Instruments/CompassCorrectionCard.xml
@@ -27,4 +27,61 @@
 			</texture>
 		</parameters>
 	</effect>
+
+	<!-- Zoom for the Card when clicked -->
+	<animation>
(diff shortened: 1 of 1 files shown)


hbeni:
@wlbragg This might also be a good solution to replace the duplicate 3D model of the zoomed card in the C172.
The advantage of the canvas solution is also, that it follows the users view as its bound to the rendering window, not the 3D model.

If you like, the cards 3D model just needs an additional pick-binding you can copy over. As we currently share the same texture, you just need to adapt the texture location to `Aircraft/c172p/Models/Interior/Panel/Instruments/mag-compass/compass-correction-card.png`, and remove the duplicate model.

PR #588, opened 2024-04-20
Add multiplayer sounds

### LINKED ISSUE #587: Add multiplayer sound, so others can hear the engine
Would be cool (and helpful) if the engine was also hearable from remote multiplayer planes.

----
Ref:
* https://wiki.flightgear.org/Howto:Add_sound_effects_to_aircraft#Multiplayer_sound
* (unanswered topic) https://forum.flightgear.org/viewtopic.php?t=17546

hbeni:
Remote C182's have engine sound now.

fix #587

### DIFF
--- a/Models/c182s.xml
+++ b/Models/c182s.xml
@@ -262,6 +262,9 @@
    ]]></load>
 </nasal>
 
+<!-- Multiplayer Sound -->
+<sound><path>Aircraft/c182s/c182-sound-multiplayer.xml</path></sound>
+
 <!-- Engine smoke -->
 <model>
   <path>Aircraft/c182s/Models/Effects/c182_enginesmoke.xml</path>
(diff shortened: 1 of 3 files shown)


hbeni:
:warning: Current state is "not working".

My sample was the 777-200 from FGAddon, and I believed that to work, but thats also broken.
However, I believe the implementation here should be "correct", but something in FGFS/next is preventing the sound from playing.
So we still need to investigate a little more.

hbeni:
pushed aa5e33970ab95333b815f5ef98166828f74fa9a6 which was working good for FGFS 2020.3.19.

With next, like said, this seems to be broken.

hbeni:
Added a SF ticket for investigation of next: https://sourceforge.net/p/flightgear/codetickets/2883/

PR #589, opened 2024-04-21
Update c182s.xml

ysopflying:
new cmm table
reference used NASA TR-1975
for 90° simple hand calculation, angles +-180 stab only alpha20-30° -> guestimate


hbeni:
Consolidated into #591

PR #590, opened 2024-04-21
Update c182s.xml

ysopflying:
CL curve corrected.


hbeni:
Consolidated into #591

PR #600, opened 2024-07-27
Use damped gnode for HeadingIndicator gyro

hbeni:
The current g-node of the directional gyro is very spike sensitive (for example taxi over hot misplaced taxilane).
This sends the gyro into tumbling mode which is not realistic.

The PR changes this, so we use a filtered g-node.

----
Related to SF ticket #2900 (https://sourceforge.net/p/flightgear/codetickets/2900/)
We need to wait for it to be merged.


hbeni:
I added an c++ internal filter to the g node, so we can rely on the instruments default behaviour now.
Should this not be enough, we can adjust the instumentation.xml `<g-filter-time>` value. And if that still should not be enough, we can calculate our own.

PR #601, opened 2024-07-27
Magnetic Compass: Switched to exposed magnetic compass props for simulating roll and pitch axis

### LINKED ISSUE #599: Magnetic Compass: Tilting wrong, should be influenced by g-forces
Some time ago, the pivot mount of the magnetic compass was added (#523).

While adding the force influences of the tilting, the reference as "earth gravity" instead of "plane gravity" which could differ.
In a coordinated bank, the compass card should not tilt like it does now, because "down" is to the floor of the plane, not earths gravity, due to centrifugal forces.

So, the tilting should be switched to the planes g-forces.
We probably can peek at the fuel system, because tank crossfeeding already uses that reference.
(edit: prop `accelerations/Ny` is used there)

@wlbragg: C172 also affected.

hbeni:
Needs SF ticket merged: https://sourceforge.net/p/flightgear/codetickets/2904/

fix #599

### DIFF
--- a/Models/Instruments/Compass.xml
+++ b/Models/Instruments/Compass.xml
@@ -34,7 +34,7 @@
 			<property>instrumentation/magnetic-compass/roll-deg-final</property>
 		</clip>
 	</expression>
-	<factor>0.5</factor>
+	<factor>1.0</factor>
 	<axis>
 		<x> 1.0 </x>
 		<y> 0.0 </y>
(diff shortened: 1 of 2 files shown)


hbeni:
:information_source: C172 also affected; when the merge in SF was done, we can tackle this.

hbeni:
Looks good when used with the FGFS ticket patches applied.
Ready for merge once FGFS/next got patched.

